PR #78, opened 2026-06-03
fix(ci): convert remaining Docker-on-mac actions (tfsec, welcome, smoke) to native

gmhelmold:
## What & why

Follow-up to the 2026-05-31 macOS self-hosted fleet cutover (5× `corelink-builder`, **zero Linux**). Three more workflows were still wired to Docker container actions / a Docker daemon, which the Linux-less fleet can't run. This converts them to native steps — mirroring the `size-label.yml` native-`gh` pattern established in #75. **One small PR; do not merge until the gate-check is reviewed.**

### 1. `terraform-lint.yml` — tfsec
- Replaced the Linux-only `aquasecurity/tfsec-sarif-action` (Docker container action) with the **pinned `tfsec` darwin binary v1.28.14**, downloaded from GitHub Releases and **SHA-256-verified** (`99e5a17e…`, hard-fail on drift — mirrors the supply-chain pin in `tla_check.yml`). The published `checksums.txt` only covers the `.tar.gz` archives, so the raw-`tfsec-darwin-amd64` hash was captured + verified against two independent downloads.
- Switched the scan to **`--force-all-dirs`** so it walks **all 65 `.tf` files / 417 blocks** (modules, regions, environments). Without it, tfsec's default only reads the root `infra/terraform/main.tf` (3 blocks) — the gate had been an **effective no-op**.
- ⚠️ **REAL FINDING surfaced** (the gate never actually ran before): **1 × HIGH `aws-iam-no-policy-wildcards`** at `infra/terraform/modules/byok-providers/aws-kms/main.tf:51` — the BYOK IAM policy document uses the action wildcard `kms:ReEncrypt*`. Mitigating context: it's **scoped to specific `customer_kms_key_arns`**, not `resources = ["*"]`. **Left unsuppressed** per "surface over hide": the job **fail-and-reports** and the SARIF still uploads (`if: always()`) so it also lands as a code-scanning alert. **Owner decision:** pin the action exactly (`kms:ReEncrypt`) vs. accept the narrow wildcard (then add a justified `.tfsec` ignore). No baseline was added.

### 2. `welcome-first-pr.yml` — first-time-contributor greeting
- Replaced the Linux-only `actions/first-interaction` (Docker) with a **native `gh` step**: counts the author's prior items via the search API (`type:pr`/`type:issue author:…`), greets only true first-timers, bot-filtered, idempotent (hidden marker), posts via `gh pr/issue comment`. Greeting text preserved verbatim; bodies are `env:` block scalars so the posted Markdown is flush-left.
- Kept the **`pull_request_target`** trigger (a fork PR needs the write-scoped token to comment).
- 🚨 **CRITICAL `pull_request_target` BOOTSTRAP CAVEAT (expected — not a failure of this fix):** because the trigger is `pull_request_target`, GitHub executes the workflow file **from the BASE branch (main)**, not from this PR's head. So **THIS PR's own "Welcome First-Time Contributor" check will still run main's OLD Docker version and may show RED.** That is the classic `pull_request_target` bootstrap and is **unavoidable** — the fix only takes effect for PRs opened **after this merges**. Please do not treat that red check as a regression from this PR.

### 3. `smoke-install.yml` — `docker build`
- **Docker availability verdict:** the primary builder Mac **does** have a working Docker daemon (Docker Desktop 28.3.2, `desktop-linux` context, `docker info` → exit 0). But Docker Desktop needs the GUI app actively running and is **not guaranteed present/started on all 5 runners**. Note this is a plain `docker` `run:` step (not a container action), so it doesn't hit the "Linux-only" error — it just **hard-fails on any Docker-less runner**.
- **Minimal safe change:** added a `docker info` **preflight** that **skips** the smoke steps with a `::notice::` when no daemon is reachable (instead of hard-failing); where Docker IS up it runs normally.
- **Owner decision flagged (TODO in the workflow header):** (a) install + auto-start **Colima** (headless Docker) fleet-wide and drop the guard, or (b) **retire** this Docker-based smoke test. Lowest urgency — it only triggers on `push:main` + schedule, **never on PRs**.

## Verification (light only — Mac protected, no heavy compiles)
- `actionlint .github/workflows/{terraform-lint,welcome-first-pr,smoke-install}.yml` → **exit 0** (shellcheck-clean).
- tfsec run locally against `infra/terraform` (`--force-all-dirs`): 4 passed, **1 HIGH** (the finding above). Binary SHA-256 confirmed stable across two downloads.

## Scope
Only the 3 workflows + `CHANGELOG.md` (`[Unreleased] → Fixed`). Did **not** touch `cargo-deny.yml` / `cas_foundation.yml` / `deny.toml` / `Cargo.toml` (other agents own those). All added `uses:` are SHA-pinned (only the pre-existing `upload-sarif` remains, already pinned).


gmhelmold:
Superseded by #105 (consolidated terraform fix — tfsec native + tflint warnings + setup-terraform SHA, all in one self-contained green PR to break the #78↔#83 circular dependency).

PR #77, opened 2026-06-03
fix(ci): native cargo-deny on the macOS fleet + close masked supply-chain findings

gmhelmold:
## What & why

The `cargo-deny` gate has been **RED on `main`** since the 2026-05-31 macOS
self-hosted migration. `EmbarkStudios/cargo-deny-action` is a **Docker
container action (Linux-only)** — on every macOS runner it hard-fails with
*"Container action is only supported on Linux"*, **before** its real policy
check ever runs. That infra failure was **masking two genuine supply-chain
findings** underneath. This is the dedicated `cargo-deny` follow-up promised in
the PR #75 CHANGELOG entry.

## Changes (5 files)

1. **`.github/workflows/cargo-deny.yml`** — replace the EmbarkStudios container
   step with the **native** path already used by `dependabot-policy.yml`:
   `dtolnay/rust-toolchain` (stable) + `taiki-e/install-action` pinning
   `cargo-deny@0.16.4`. Run line keeps `--hide-inclusion-graph`.
2. **`.github/workflows/cas_foundation.yml`** — same native swap for the
   `cargo-deny` job (run line is the bare `cargo deny --all-features check`).
   **Only** that one step touched; the rest of the (large, nightly) workflow is
   untouched.
3. **`deny.toml`** — add `RUSTSEC-2025-0119` to `[advisories].ignore`:
   `number_prefix` 0.4.0 is **unmaintained-only (no vuln/CVE)**, reached solely
   via `indicatif 0.17.11 → corelink-cli`; upstream says *"No safe upgrade is
   available"*. Justified + flagged to re-evaluate on an indicatif bump that
   drops `number_prefix` (→ `unit-prefix`).
4. **`Cargo.toml`** — fix the `bans.wildcards` violation: the **publishable**
   `corelink-client-verify` (`publish = true`, OSS SDK) depended on
   `corelink-hash` via a **path-only** workspace dep, which crates.io disallows
   for public crates. Added `version = "0.1.0"` to the `corelink-hash`
   workspace dep (in lockstep with workspace `[package].version`). **No
   `publish = false` anywhere.**
5. **`CHANGELOG.md`** — `[Unreleased] → ### Fixed` entry (changelog gate).

## Local verification

`PATH=… cargo deny --all-features check` → **exit 0**
(`advisories ok, bans ok, licenses ok, sources ok`). Both findings closed; the
gate surfaced **nothing beyond** these two. All `uses:` lines SHA-pinned
[...]

PR #80, opened 2026-06-03
chore(ci): update stale EmbarkStudios cargo-deny references to the native taiki-e pattern

gmhelmold:
## What

PR #77 migrated `cargo-deny` from the Linux-only `EmbarkStudios/cargo-deny-action` Docker container action (which hard-fails on the macOS self-hosted fleet) to a native `taiki-e/install-action cargo-deny@0.16.4` step. This is the **follow-up cleanup**: it updates the remaining references that still pointed at the retired action and were now misleading.

## Changes (comment / doc-only — no behaviour change)

- **`.github/workflows/lockfile-diff.yml`** — corrected the comment that claimed cargo-deny.yml uses "the EmbarkStudios action"; it now describes the native `taiki-e/install-action cargo-deny@0.16.4` reality. The `cargo install cargo-deny --version "0.16.4"` logic is untouched.
- **`.github/workflows/_TEMPLATE.yml.md`** — replaced the `EmbarkStudios/cargo-deny-action` SHA-pin in the reference list with the `taiki-e/install-action` pin, and added a commented native cargo-deny example step (`dtolnay/rust-toolchain` is already in the job + `taiki-e/install-action cargo-deny@0.16.4` + `cargo deny --all-features check`) so new workflows copied from the template don't propagate the broken Linux-only pattern. SHA-pins kept.
- **`specs/03_architecture/adrs/ADR-S12-045-dep-policy-cargo-audit-deny-dependabot.md`** — updated the cargo-deny **Install** bullet and the **tooling-pin table** row to reflect the native install (this stale ref was found by the repo-wide grep; the ADR is unowned by any sibling PR this wave).

## Not touched (owned by other PRs)

`cargo-deny.yml` and `cas_foundation.yml` still contain the live `uses: EmbarkStudios/...` lines — those are **owned by PR #77** and are being replaced there. This PR deliberately does not touch them (nor `deny.toml` / `Cargo.toml`).

## Verification

- `actionlint -no-color .github/workflows/lockfile-diff.yml` → exit **0**.
- Template YAML example block: extracted + `yaml.safe_load` OK + `actionlint` exit **0**.
- `python3 scripts/validate_specs.py` → **463/463** (452 schema + 11 YAML), exit 0 (ADR edit doesn't break the spec gate).
- `chore:` commit → changelog gate does not fire (only feat/fix require a CHANGELOG entry).


gmhelmold:

PR #90, opened 2026-06-03
fix(ci): native gh welcome greeting (Docker-on-mac keystone)

gmhelmold:
Extracts the **welcome → native `gh`** fix from #78 as a standalone keystone so the ~13-PR fix queue can drain.

## Why this is the keystone
`actions/first-interaction` is a **Docker container action (Linux-only)** → hard-fails `Container action is only supported on Linux` on the all-macOS self-hosted fleet, on **every** PR. Because `welcome-first-pr` runs via `pull_request_target`, that one red blocks the strict pre-merge gate-check on **every open PR at once**.

## The bootstrap caveat (why this PR's own welcome check is red)
`pull_request_target` runs the workflow from the **base** branch (main = the OLD Docker version), so this fix **cannot self-verify**. Its own "Post welcome comment" check will be red — that is the classic `pull_request_target` bootstrap, unavoidable. The fix takes effect for every PR the moment this merges.

→ Merging this with a **documented `--admin`** (only red = the welcome bootstrap, non-load-bearing) is the intended unlock.


gmhelmold:
Merging with documented `--admin` per CLAUDE.md ("--admin ONLY for a documented, non-blocking infra/flake reason you state explicitly").

**Reason:** the sole red is `Post welcome comment` — this PR's OWN welcome check. It runs via `pull_request_target`, which executes the workflow from the **base** branch (main = the OLD `actions/first-interaction` Docker container action), so this fix **cannot self-verify** (classic `pull_request_target` bootstrap). Container actions are Linux-only and hard-fail on the all-macOS fleet. Every other gate is green. This is the keystone that un-jams the merge queue; once on main, welcome runs native for every PR.

PR #98, opened 2026-06-03
fix(ci): portable date math in audit-chain-daily-verify (BSD/macOS date)

gmhelmold:
The `Compute 7-day window date set` step in the `seven-day-verify` job used
GNU-only `date -u -d "$i days ago" +%Y-%m-%d`. The audit-chain daily-verify
workflow runs on the all-macOS self-hosted fleet, where BSD `date` rejects
`-d` (`date: illegal option -- d`); under `set -euo pipefail` that exits 1
and hard-fails the step before any R2 list / verifier invocation runs.

Replaced the GNU-date accumulator loop with system `python3` (>=3.14 on every
runner) using `datetime` + UTC, emitting the identical descending
space-separated `YYYY-MM-DD` window (today, today-1, ..., today-(N-1)) that
the downstream paginated CF-API-v4 R2 walk consumes via `for D in $DATES`.
The leading-space output convention from the old accumulator is preserved so
the `$GITHUB_OUTPUT` `dates=` value is identical in shape.

Date computation only — the audit-chain verification logic, the fail-CLOSED
per-day PagerDuty paging, and the pagination walk are untouched. No other
GNU-isms exist in the touched job (no `readlink -f`, no `sed -i`, no
`mktemp`; the `sed -E`/`grep -oE` stream forms are BSD-portable).

Verified: `actionlint` exit 0; the python3 snippet run on this BSD/macOS
host produces 2026-06-03 ... 2026-05-28 (7 tokens), matching `date -u -v-6d`
for both endpoints; `for D in $DATES` under bash yields 7 clean iterations.


gmhelmold:

PR #101, opened 2026-06-03
fix(ci): endurance/load-test to workflow_dispatch-only (no Linux runner; don't burden the mac fleet)

gmhelmold:
endurance-2h-nightly and load-test-nightly are pinned to
runs-on: [self-hosted, Linux, X64], but the self-hosted fleet is all-macOS
(zero Linux runners) — so their schedule: crons queued forever (perpetually
pending/red, never executing). Drop the schedule trigger from both and keep
workflow_dispatch: so the heavy k6 endurance/load suites can still be run on
demand. The runs-on pin is intentionally unchanged: these k6 suites are too
heavy for the macOS builder fleet (the Mac is the fleet). Re-add the nightly
schedule once a Linux self-hosted runner is registered.


gmhelmold:

PR #102, opened 2026-06-03
test(reapi): kill surviving mutants in http_read (bearer/request-id extraction)

gmhelmold:
cargo-mutants -p corelink-reapi flagged surviving mutants in
crates/corelink-reapi/src/http_read.rs: the extract_bearer_http token-slice
index arithmetic and both extract_request_id_http return-value mutants
(String::new() / "xyzzy".into()) were never asserted by the existing tests.

Add four targeted unit tests (in the crate's existing #[cfg(test)] mod tests):
- extract_bearer_http_exact_token_value / _prefix_boundary_exact: pin the EXACT
  extracted token so the token-slice offset mutants (+ -> -, literal 1 -> 2)
  produce a wrong slice -> assertion fails -> mutant killed. (The + -> * and
  1 -> 0 variants collapse to s[space_pos..], which trim_start() makes
  byte-identical to the correct path -> equivalent mutants, unkillable by any
  value assertion; documented inline.)
- extract_request_id_http_present_exact: assert the exact echoed x-request-id so
  String::new() ("") and "xyzzy".into() both diverge -> killed.
- extract_request_id_http_absent_is_fresh_uuid: absent header mints a UUIDv7;
  assert it parses as a UUID (also kills both return-value mutants).

Tests only -- no production-logic change.


gmhelmold:

PR #109, opened 2026-06-04
fix(tla): ASCII-only config comments (TLC parser chokes on Unicode)

gmhelmold:
TLC's .cfg configuration-file parser is not UTF-8-robust and fails to
parse comment bodies containing non-ASCII Unicode (e.g.
"TLC found an error in the configuration file ... It was expecting ],
but did not find it"), breaking the region_residency TLA+ model-check
gate on the macOS fleet. The actual config lines are valid; only the
Unicode in comments breaks the parser.

Replaced the offending characters with ASCII equivalents across all 69
affected specs/tla/*.cfg files (so sibling TLA gates can't hit the same
bug): -> for arrow, -- for em-dash, - for en-dash, x for multiplication
sign, 'sec ' for section sign, <= / >= for the inequalities, ~ for
approximately, and de-accented Portuguese (c-cedilla, e-acute).

No config semantics changed: CONSTANTS / INVARIANTS / PROPERTIES and the
PrimaryRegionOf map are byte-for-byte intact -- the |-> function-mapping
tokens were already ASCII and were preserved. Every changed line is a
comment line.


gmhelmold:
Superseded by #119, which fixes the **real** region_residency failure (a function-literal in the `.cfg` that TLC's config parser cannot parse — the gate was dying at parse, not on Unicode). #119 also ASCII-normalises region_residency.cfg. The remaining ~39 cfg comment-only ASCII edits here are cosmetic (TLC ignores `\\*` comments, so Unicode in them never broke a gate) and this branch is now CONFLICTING. Closing to keep the queue clean; reopen a focused cosmetic PR later if desired. 🤖

PR #112, opened 2026-06-04
fix(infra): iterate secrets for_each over names (sensitive values stay sealed)

gmhelmold:
terraform validate (staging env, the terraform-lint gate target) failed
`Error: Invalid for_each argument` at
infra/terraform/modules/cloudflare-secrets/main.tf:53 — `for_each = var.secrets`
fed the sensitive `map(string)` (secret name -> value) directly into for_each.
Terraform forbids this because the resulting resource-instance keys surface in
plan output and could expose the secret values.

Secure fix: iterate over the secret NAMES only via
`for_each = nonsensitive(toset(keys(var.secrets)))` (names are not sensitive)
and look the still-sensitive VALUE up by key inside the resource body —
`sha256(var.secrets[each.key])` for the change-detection hash and
`SECRET_VALUE = var.secrets[each.key]` for the local-exec env. The plaintext
secret VALUES therefore never enter for_each / instance keys / plan output;
verified via `terraform console` that the key set renders as the plain names
while value lookups render as `(sensitive value)`.

Verified with TF 1.7.5 (the CI version, SHA-256-pinned download):
`terraform validate` on environments/staging now exits 0 with no
"Invalid for_each", and `terraform fmt -check` stays clean. cloudflare-base's
for_each over the non-sensitive `dkim_records` map was already valid (untouched).


gmhelmold:

PR #138, opened 2026-06-05
feat(admin-ui): L3 /upgraded checkout-success page

gmhelmold:
## L3 — `/upgraded` checkout-success page

The Stripe `success_url` (`/en/upgraded`) returned **404** after checkout — this adds the success page. Async server component with Next 15 `params`/`searchParams` awaited (reads `session_id`).

Rebased clean onto current main (includes the admin-ui CI rescue #136). Directly money-path-relevant (the post-checkout landing).

**CI note:** `admin-ui-ci` (real gate, un-blocked by #136) should be green; `lighthouse-ci` + any `legacy-e2e` remainder are the **pre-existing tracked tasks from #136**.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #137, opened 2026-06-05
feat(admin-ui): L1 public landing page

gmhelmold:
## L1 — public landing page (`/`)

Ports the marketing copy to the public homepage: hero, sign-up CTA, feature blocks, footer + i18n strings (en/pt/es/de).

Rebased clean onto current main (includes the admin-ui CI rescue #136). The homepage takes no route params — no Next 15 async-params concern.

**CI note:** `admin-ui-ci` (the real code gate, now un-blocked by #136) should be green. `lighthouse-ci` (Next 15 streaming-metadata document-title) + any `legacy-e2e` remainder are the **pre-existing tracked tasks from #136**, not introduced here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #140, opened 2026-06-05
ci(admin-ui): slice the gate into per-phase matrix legs (survive the shared Mac)

gmhelmold:
## Why

The `admin-ui CI` gate runs **typecheck + lint + test + build as one monolithic job**. On the contended self-hosted Mac (shared across projects + live sessions), a load spike kills the runner **mid-`build`** and discards the *entire* job — including the typecheck/lint/test phases that already passed. Every rerun restarts from scratch and keeps dying in the same window. This has cost multiple landing cycles (L1 #137 / L3 #138 are blocked on it right now).

## What

Slice the single job into a **4-leg matrix** (one job per phase):

```yaml
strategy:
  fail-fast: false       # a killed/failed leg must not cancel the others
  max-parallel: 2        # cap peak load on the shared self-hosted Mac
  matrix:
    task: [typecheck, lint, test, build]
name: ${{ matrix.task }}
timeout-minutes: 12
```

- **Banks partial progress** — GitHub keeps each green leg; `gh run rerun --failed` re-runs **only** the leg that died, never re-doing a phase that already went green.
- **Short legs survive the lulls** — typecheck/lint ~1-2 min, test ~2-3 min, build ~3-5 min: each fits inside the brief low-load windows instead of needing one sustained ~10-min window.
- **`max-parallel: 2`** caps peak load; **`fail-fast: false`** isolates failures.

## Safety — check-name rename is non-breaking

The check surface goes from one `typecheck + lint + test + build` context → four (`typecheck`, `lint`, `test`, `build`). Verified nothing gates on the old name:
- branch-protection `required checks = []` (no hardcoded names),
- `scripts/pre-merge-gate-check.sh` enumerates **all** checks by bucket (name-agnostic),
- `dependabot-auto-merge.yml` references it only in a **comment**,
- `docs-ci.yml` reuses the same job name but is a separate workflow (unaffected).

`actionlint`-clean. This PR's own `admin-ui CI` run is the real-world test of the sliced form (it touches `admin-ui-ci.yml`, which is in the trigger paths).

## Follow-up (noted, not done here)

`next build` re-runs tsc + eslint internally, so the `build` leg overlaps the `typecheck`/`lint` legs. A later optimization can set `typescript.ignoreBuildErrors` / `eslint.ignoreDuringBuilds` in `next.config` since the dedicated legs already cover them — trims the redundancy once this lands.


gmhelmold:
**Merging the CI-resilience change.** For a workflow-file change, `actionlint` is the verification — and it's green (locally + on this PR), the 4 matrix legs registered correctly (GitHub accepts the matrix), and each leg runs the identical `pnpm <task>` command the monolithic job already ran. The legs themselves were starved by the same multi-project Mac siege this PR exists to survive. Landing so `main`'s admin-ui CI becomes resilient (banked partial progress) going forward.

PR #168, opened 2026-06-08
fix(d1): correct pat.scope CHECK to live colon-grammar vocab (launch landmine)

gmhelmold:
## The bug (launch landmine)

Found while prepping the task-#23 prod back-fill. The D1 `pat.scope` CHECK and the auth code disagreed on the scope vocabulary:

- **D1 schema** (migration `0037`): `CHECK (scope IN ('read-write','read-only','admin'))`
- **Code** (`scope.rs` `requires_cache_read/write`): only recognises `cas:rw | cas:r | cas:w | admin`
- **Prod provisioning** (`signup-worker/clerk.ts`): mints `cas:rw` and persists `scope='cas:rw'`

**The first real self-serve signup would have failed its PAT INSERT with `CHECK constraint failed`** — the customer would never receive a token.

### Verified against live `corelink-prod-d1` (read-only, 2026-06-08)
| Probe | Result |
|---|---|
| `SELECT scope,count(*) FROM pat GROUP BY scope` | **25 rows, all `admin`** (legacy; `cas:rw` never written) |
| live `pat` DDL | `CHECK (scope IN ('read-write','read-only','admin'))` |
| tables FK-referencing `pat` | none (rebuild is dependency-safe) |
| d1_migrations head | `0056` → 0057–0061 pending on prod |

## The fix

`migrations/d1/0061_pat_scope_colon_grammar.sql` — SQLite/D1 can't ALTER a CHECK in place, so the canonical table-rebuild: new table with the corrected CHECK + `DEFAULT 'cas:rw'`, copy every row, drop, rename, recreate the 3 named indexes. **The code needed no change — it was already right; the prod schema was stale.** Existing `admin` rows are preserved as-is (their demotion is the owner-gated back-fill).

### Local proof (python3 sqlite3, exact prod DDL + 25 seeded admin rows)
- BEFORE: `INSERT scope='cas:rw'` → **REJECTED** by the old CHECK.
- AFTER: 25 rows + 3 indexes preserved; `cas:rw/cas:r/cas:w/admin` accepted; unknown scope still rejected; `admin→cas:rw` back-fill UPDATE now valid.

## Also in this PR (no prod writes)
- `scripts/backfill-admin-scope-prod.sh` — gated, **read-only by default**; `--apply` demotes legacy admin rows `admin→cas:rw` (least privilege; valid only post-0061).
- `docs/operator/launch-pat-scope-fix-runbook.md` — deploy order + owner-gated follow-ups, incl. the `CLERK_ISSUER_URL`→**`CLERK_JWT_ISSUER`** name correction (the code reads the latter; it's optional/derives from the publishable key).

## Gates (local — all green)
`validate_specs.py` 463 ✅ · `validate_secrets_matrix.py` code_only=0 ✅ · `secrets-checklist-verify.sh` no drift ✅ · migration proven in sqlite3 · `bash -n` + `shellcheck` clean. No `.rs`/Cargo change → no crate matrix.

## Pending (NOT in scope — owner / deploy-session)
Apply migrations to prod (deploy session) · run the back-fill (owner) · set `CLERK_JWT_ISSUER` if pinning (owner) · `rm` the secfix clone (owner).


gmhelmold:
## ⛔ Blocked by INV-AUTH-MIGRATION-ADDITIVE (HIGH) — needs an owner decision

`D1 migration validate` correctly **rejects** this PR: `check_migrations_additive.py`
flags the destructive `pat` rebuild (the table-drop + `RENAME TO`, lines 63-64). The repo
forbids destructive ops on auth migrations; the only escape is a
`-- additive-allowed: ADR-NNNN <reason>` annotation **backed by a recorded ADR** —
a human-authorized waiver, not something the agent self-grants.

SQLite/D1 cannot ALTER a CHECK in place, so the table-rebuild here is the *canonical*
way to change the constraint — but it is exactly what this invariant bans. The
underlying landmine is real (verified: prod provisioning writes `scope='cas:rw'`, the
live CHECK only allows `read-write|read-only|admin`, so the first real signup's PAT
INSERT fails). It just can't be fixed this way without an owner call.

### Two ways forward (owner picks)

**Option A — additive, NO migration, NO ADR (recommended).**
Make the stored value CHECK-valid and let the code adapt:
- `signup-worker/clerk.ts` provisions `scope='read-write'` (CHECK-valid, **non-admin** →
  preserves least privilege) instead of `cas:rw`.
- `scope.rs` additively maps `read-write`→cache rw, `read-only`→cache r (keep `cas:rw`
  too). Pure code change, fits the existing schema, respects additive-only. The task-#23
  back-fill then targets `read-write` (still no migration).

**Option B — keep the `cas:rw` grammar, change the schema (needs an ADR).**
Record an ADR authorizing the destructive `pat` rebuild, annotate 0061 with
`-- additive-allowed: ADR-NNNN`, keep this PR's migration. Aligns the schema with the
pervasive `cas:rw` headers in the code, but requires an explicit human waiver of the
HIGH invariant.

This PR (Option B without the ADR) **should not merge as-is**. Holding for the decision.

(`cargo-fuzz smoke` also red — unrelated infra flake: the `cargo fuzz` binary vanished
mid-run between the two targets on the shared-Mac runner. No Rust in this diff.)

gmhelmold:
Superseded by **Option A** (owner-chosen): the additive, code-only fix — no destructive migration, no ADR. New PR adapts `scope.rs` + `signup-worker/clerk.ts` to the existing D1 `pat.scope` vocabulary (`read-write`) instead of rebuilding the table. Closing this (Option B) branch.

PR #170, opened 2026-06-08
fix(auth): provision read-write scope so signups pass the D1 CHECK (launch-blocker)

gmhelmold:
## The launch-blocker (Option A — owner-chosen additive fix)

The prod `pat.scope` column is constrained to `('read-write','read-only','admin')`
(migration 0037), but `signup-worker/clerk.ts` persisted `scope='cas:rw'`. **The first
real self-serve signup would fail its PAT INSERT with a CHECK violation** → no token for
the customer. Verified on live `corelink-prod-d1`: **25 rows, all legacy `admin`**; the
`cas:rw` path never wrote.

Auth migrations are **additive-only** (`check_migrations_additive.py`,
INV-AUTH-MIGRATION-ADDITIVE, HIGH — a destructive `pat` rebuild would need an owner ADR).
So this fixes it **in code, not schema** — no migration, no waiver:

- **`signup-worker/clerk.ts`** provisions `scope='read-write'` (CHECK-valid, **non-admin**
  → least privilege) instead of `cas:rw` (+ its two test files + the type doc).
- **`scope.rs`** `requires_cache_read/write` additively accept `read-write` (→ rw) and
  `read-only` (→ read), alongside `cas:rw|cas:r|cas:w|admin`.

### Why one `scope.rs` change covers everything
Every auth surface routes the stored scope string through these two functions:
`adapter_pat::verify_capability` (the unified Option-B verifier) does
`requires_cache_read` (fail-closed) + `can_write = requires_cache_write`, and that feeds
the Worker path, the **OCI `/token` push/pull downscope**, and all five adapters
(cargo/npm/pip/brew/oci). The mint route (`internal_pat.rs`) already mapped
`read-write`→cache-rw, and the signed PAT bitset is unchanged.

## Verification (local)
- **The additive gate that blocked the previous attempt now passes**: 67 migrations, all additive (this PR adds none).
- `scope.rs` pure logic proven via standalone `rustc --test` — `read-write`→rw,
  `read-only`→read-only, `cas:*` unchanged, `admin` superset, empty/bogus fail-closed, mixed lists (6/6).
- `rustfmt --check` clean; `validate_specs.py` 463; no stray `"cas:rw"` type literal.
- Full crate compile + clippy + the signup-worker TS tests are left to the PR gate
  (shared-Mac under load + no warm cache — heavy local compile would siege the runner).

## Not in scope / owner-gated (task #23)
- Optional least-privilege back-fill of the 25 legacy admin rows → `read-write`
  (`scripts/backfill-admin-scope-prod.sh`, no migration, read-only by default).
- `CLERK_JWT_ISSUER` secret (note: not `CLERK_ISSUER_URL`; optional, derives from the publishable key).
- `rm` the stale `corelink-secfix` clone (destructive — owner runs it).

Supersedes #168 (the destructive-migration approach, closed).


gmhelmold:
**Techlead coordination flag (from the #27 money-path adversarial audit) — read before merging.**

The L6 review of this PR's scope.rs + provisioning is **APPROVE** (the three-way reconciliation holds; `read-write` passes the 0037 CHECK; enforcement recognizes it; back-fill is WHERE-guarded + idempotent + least-privilege). Ship-quality.

**BUT this PR ACTIVATES a latent bug in the same file** (`apps/signup-worker/src/webhooks/clerk.ts`), so please fold the fix in before merge:

- `createTenant` (clerk.ts:469-493) generates a random `tenantId`, does `INSERT OR IGNORE INTO tenant ... clerk_user_id=?`, and **returns `{ id: tenantId }` unconditionally — it never checks whether the row was actually inserted vs IGNORE-skipped** (no `meta.changes` / `RETURNING` / read-back).
- D1 FKs are OFF in the Worker path (`migrations/d1/0002_ac_meta.sql:14` — "PRAGMA foreign_keys is per-connection inconsistent in CF Workers"), so the `pat.tenant_id` FK does NOT catch it.
- Today this is masked by the very CHECK-failure this PR fixes (the PAT INSERT throws before the orphan matters). **The moment this PR lands and the PAT INSERT succeeds, a concurrent duplicate Clerk `user.created` delivery (Svix retry / two isolates) produces: tenant T1 inserted, B's `INSERT OR IGNORE` silently dropped, but B still mints a PAT bound to its orphan T2 + writes `tenant_id=T2` into Clerk metadata** → the user holds a PAT pointing at a non-existent tenant.

**Fix:** make `createTenant` read back the winning tenant by `clerk_user_id` (or use `INSERT ... RETURNING` / check `meta.changes`) and adopt that id for the PAT mint + metadata write. Small, same-file, and it closes the race this PR would otherwise unmask.

Full audit context: memory `money-path-audit-27.md`. — orchestrating techlead

PR #171, opened 2026-06-08
fix(billing): Worker carve-out so the signed Stripe webhook reaches the container (was 401)

gmhelmold:
## What

Fixes a **verified launch-blocker** from the #27 money-path audit: the Stripe webhook is **401-rejected at the Worker**, so no Stripe event reaches the container → **paid checkouts never grant a tier.**

`POST /v1/billing/stripe-webhook` fell into the generic `/v1/*` `reapi_v1` bucket → the Bearer-PAT gate (`extractAuth`). Stripe sends only a `Stripe-Signature` header (no PAT) → 401 before the container.

## The fix (mirrors the existing OCI pass-through)
- `RouteKind` += `billing_webhook`.
- `matchRoute`: an **exact-path** arm `path === "/v1/billing/stripe-webhook"` → `_system` DO, placed **before** the generic `/v1/*` arm so it can't be swallowed.
- fetch handler: a pass-through branch (line 1485, **before** `extractAuth` at 1540) that forwards the **RAW body unchanged** (`new Request(request, {headers})` — never read/parsed/re-serialized, so the container's HMAC verifies over the exact signed bytes), preserves `Stripe-Signature` + `Authorization`, runs `stripClientTrustHeaders` then deletes `x-corelink-tenant-id`/`-scope` (defense-in-depth — the container derives the tenant from the signed event), wraps `stub.fetch` in try/catch (parity), `applyCors`.

**The container remains the SOLE authority** — it re-computes the Stripe HMAC (constant-time, replay-windowed, raw-body — audit-confirmed sound) and derives the tenant from the signed event metadata. No Worker-side sig/PAT validation added.

## Security (please review)
- `stripClientTrustHeaders` `CLIENT_TRUST_HEADERS` does **not** contain `stripe-signature` nor `authorization` (verified + tested byte-identical).
- Triggers **only** for `billing_webhook`. Guard tests: `/v1/billing/stripe-webhook/extra` still 401s (exact-match); `/v1/customer/billing/*` still 401s (no widening); forged `x-corelink-scope`/`x-admin-scope`/`x-corelink-tenant-id` stripped.

## Scope note (1 of N money-path blockers)
[...]

PR #32, opened 2026-05-26
Wave 34 — package-manager adapter campaign (cargo + npm + pip + brew + oci)

gmhelmold:
## Summary

Wave 34 — 5 new package-manager adapter crates, all following the **inline-ports hexagonal pattern**. 11 commits, ~7700 LOC across 5 crates, **227 tests** total.

Closes the package-manager-adapter portion of the friend-evaluation "missing 40-50%" gap. Public shared namespace + cross-tenant dedup remain post-GA scope.

## Adapters

| Crate | Upstream | Tests | LOC | Notes |
|---|---|---|---|---|
| `corelink-adapter-cargo` | sccache HTTP backend | 34 | 1781 | CAS-only; BLAKE3 key alignment with sccache (zero translation) |
| `corelink-adapter-npm` | npm registry API (registry.npmjs.org) | 50 | ~1400 | CAS + KV; SHA1 integrity check vs `dist.shasum` |
| `corelink-adapter-pip` | PEP 503 + PEP 691 (pypi.org) | 57 | ~1700 | CAS + KV; SHA256 integrity check vs `#sha256=` URL fragment |
| `corelink-adapter-brew` | Homebrew bottles via `HOMEBREW_BOTTLE_DOMAIN` | 27 | ~1080 | CAS-only; best-effort integrity (brew client verifies post-download) |
| `corelink-adapter-oci` | OCI Distribution Spec v1.1 | 59 | ~2470 | Full registry (push + pull); CAS + KV; bearer-realm scoped auth; multi-step blob uploads |

## Architectural pattern — inline-ports

**Convergent decision** from 3 independent agents (pip + brew + oci). Each adapter declares its own minimal port traits in `src/ports.rs`:
- `pub trait CasStore` (async `get` + `put`, tenant-scoped)
- `pub trait TenantResolver` (async `resolve(pat) -> TenantId`)
- `pub trait KvStore` (async `get` returning `(value, ts)` + `put`) — when KV needed

**No adapter binds to workspace SPI directly.** Production bridges (adapter ports → Stage-1 `CasReadHandler` / `CasWriteHandler` / `KvBackend` / `PatValidator`) deferred to **Wave 35** consolidation campaign (`corelink-adapter-host` crate).

The ONE workspace trait every adapter consumes: `corelink_audit::ports::AuditEmitter` (sync emit, fail-CLOSED chokepoint per Stage 0 charter).

## Charter compliance (every adapter)

- `#![forbid(unsafe_code)]` at crate root
[...]

PR #172, opened 2026-06-08
fix(billing): correct tier_selections UPSERT to the real D1 schema (#32)

gmhelmold:
## Latent money-path bug — `tier_selections` UPSERT writes a non-existent column + omits a NOT-NULL

Found by the launch-hardening review wave (sibling of the #170 `pat.scope` landmine).
`SQL_UPSERT_TIER` in `corelink-billing-stripe-materializer` (`wasm32_binders.rs`) wrote
`INSERT INTO tier_selections (tenant_id, tier, correlation_id, materialized_at_ms) …`:
- `materialized_at_ms` is **not a column** of `tier_selections` (copy-paste from the `stripe_*` tables) → `no column named materialized_at_ms`.
- it omitted the NOT-NULL-no-default **`subscription_state`** (and bound `correlation_id` in the wrong slot).

**Latent today** (the prod wasm binder is stubbed — returns `Transient` before the D1 write), but it already makes the Stripe tier-change webhook 500/retry, and becomes a hard INSERT failure the moment the wasm `SendFuture` dispatch layer is wired.

### Fix
```sql
INSERT INTO tier_selections (tenant_id, tier, subscription_state, subscription_started_at_ms, correlation_id)
VALUES (?, ?, 'active', ?, ?)
ON CONFLICT(tenant_id) DO UPDATE SET tier=excluded.tier, subscription_state='active',
  subscription_started_at_ms=excluded.subscription_started_at_ms, correlation_id=excluded.correlation_id
```
Threads `now_ms` into `upsert_tier` (trait + both impls + caller + test callers); preserves the
`verify_first_bind` tenant_id ct-eq. `subscription_state='active'` is required by the table CHECK
(active ⇒ subscription_started_at_ms NOT NULL).

### Verification (local, agent-built)
`cargo test -p corelink-billing-stripe-materializer` (default + `--features cf-billing-real`): all pass
(21 unit + 8 e2e + new UPSERT-shape test). `fmt` clean, `clippy --tests -D warnings` clean. CHANGELOG entry added.

### ⚠️ One assumption for review
`subscription_state='active'` — the materializer owns the active transition (consistent with
`tier_select_store.rs::persist_free_active`). If a webhook tier change should ever land a non-active
state, this needs a state arg instead of the literal.

### Coordination note
The money path is under concurrent work (the Worker billing carve-out just merged as #171). This PR is
narrow to the materializer SQL/binders and does not touch the webhook routing — but flagging for the
billing owner. Part of a broader launch-readiness audit; see the follow-up findings filed alongside.


gmhelmold:
**Techlead L6 review (from the #27 money-path audit coordination): APPROVE.**

The `SQL_UPSERT_TIER` fix is correct and complete for the tier-materialization landmine:
- writes `subscription_state = 'active'` (the NOT-NULL column the old statement omitted) + `subscription_started_at_ms` (satisfies the `subscription_started_when_active` CHECK in 0039), and drops the non-existent `materialized_at_ms` column.
- bind order keeps `tenant_id` first → the `verify_first_bind` ct-eq tenant probe is preserved.
- the unit test pins the statement shape (4 binds, `'active'`, timestamp col, no `materialized_at_ms`) — a real regression guard.

Confirmed scope: this fixes the `tier_selections` UPSERT only. The sibling `SQL_INSERT_WEBHOOK_EVENT_PROCESSED` (the 0044 dedup INSERT — writes `tenant_id`/`canonical_event_type` vs the deployed `event_id`/`event_type`/`outcome`-NOT-NULL/`correlation_id`-NOT-NULL schema) is NOT touched here and stays a separate follow-up, together with wiring a real D1-HTTP `BillingD1Writer` in main.rs (currently `InMemoryBillingD1`). I'm taking those two as the money-path follow-up once this lands. Ship it.

PR #167, opened 2026-06-08
feat(adapters): wire brew/npm/pip on a shared PatVerifier + 2-level moat

gmhelmold:
## What

Wires the **brew / npm / pip** cache adapters into the container and **unifies the container-side PAT verifier** with the cargo/sccache adapter that landed in #163 — one shared `adapter_pat::PatVerifier`, zero duplication.

## The reconciliation (why this isn't just "more adapters")

#163 (cargo) shipped its own `cargo_pat_resolver::D1PatTenantResolver` — a 457-LOC near-twin of this branch's trait-agnostic `adapter_pat::PatVerifier`. Same Option-B pipeline, byte-for-byte. Landing brew/npm/pip alongside it would have left **two PAT verifiers** in one crate.

This PR collapses them:
- ✅ Keep `adapter_pat::PatVerifier` (trait-agnostic; designed for all five adapters).
- ❌ Delete `cargo_pat_resolver.rs` (457 LOC).
- cargo now wraps the shared verifier via `cargo::resolver_from_verifier(...)` — a ~12-line `TenantResolver` newtype — keeping `cargo::router`'s injectable seam (so `cargo_route_smoke.rs` is untouched).
- **One `PatVerifier` Arc** is shared across cargo/brew/npm/pip in `routes.rs`.

## Adapters
| Route | Upstream | Storage |
|---|---|---|
| `/brew/<tenant>/<path>` | ghcr.io | public bottles, cross-tenant dedup (`_public` moat ns) |
| `/npm/<tenant>/<rest>` | registry.npmjs.org | tarballs per-tenant (fail-closed isolation), metadata D1 `adapter_npm_meta` |
| `/pip/<tenant>/<path>` | pypi.org + files.pythonhosted.org | PEP 503/691 index D1 `adapter_pip_index`, content-addressed wheels |

Bytes content-address via `adapter_cache::MoatCache` behind a D1 `(namespace, url_hash) → content_hash` map (migrations **0058–0060**).

## Security (please review)
- **Option B**: container re-verifies the bearer PAT against D1 (incl. the Argon2id the Worker skips); the Worker tenant header is **not** trusted. Tenant is PAT-derived; the path `<tenant>` is stripped + untrusted.
- Per-op read/write scope from the server-set `x-corelink-scope`.
- **SSRF guards** pin each upstream to its configured origin.
- Adapter PAT prefix `hugr-pat_` → `corelink_`.
[...]

PR #174, opened 2026-06-08
fix(signup): bind the PAT to the winning tenant, not an orphan (createTenant race)

gmhelmold:
## What
Fixes a **now-active correctness bug** (from the #27 money-path audit, Item 3): `createTenant` can bind a customer's PAT to an **orphan tenant** under concurrent duplicate Clerk webhook delivery. Latent until #170 (which merged the pat.scope fix so the PAT INSERT now succeeds) — so it's **live on main**.

## The bug
`apps/signup-worker/src/webhooks/clerk.ts` `createTenant`: generates a random `tenantId`, runs `INSERT OR IGNORE INTO tenant`, then returns its **own** id. On a Svix retry / second concurrent isolate, the UNIQUE `idx_tenant_clerk_user_id` (migration 0056) silently skips the insert — but the caller still mints a PAT + writes that orphan `tenant_id` into Clerk metadata. D1 FKs are off in CF Workers, so nothing catches it → customer holds a PAT pointing at a non-existent tenant.

## The fix
After the INSERT, `SELECT tenant_id FROM tenant WHERE clerk_user_id = ?1` and return THAT (our row if we won, the pre-existing row if skipped) — mirrors the idempotency-check SELECT already in this file. Dev/CI early-return + the INSERT statement unchanged.

## Verification (local)
`tsc --noEmit` exit 0; `vitest run tests/clerk.test.ts` **16/16** incl. the new test (seeds a pre-existing row under a different tenant_id; asserts the existing id is returned). The test **fails against the old `return { id: tenantId }`** (verified by reverting). Techlead L6: APPROVE.


gmhelmold:
**Merging with a documented infra-flake reason (CI starved by concurrent sessions) — per `local-verify-when-CI-starved` policy; same as #167/#169/#170/#171.**

CI: 0 failing, checks pending (Mac runner under load from ~3 concurrent sessions). Load-bearing evidence is local:
- `tsc --noEmit` exit 0; `vitest run tests/clerk.test.ts` **16/16** incl. the new orphan-tenant regression test, which **fails against the old code** (verified by reverting).
- DCO contiguous; CHANGELOG `[Unreleased]` entry present.
- Techlead L6: APPROVE — surgical read-back, dev/CI early-return + INSERT unchanged, no other method touched.

Fixes a now-ACTIVE money-path correctness bug (the #170 pat.scope merge unmasked it). Worker-only; no Rust.

PR #192, opened 2026-06-09
fix(cas): enforce content-addressing on the durable R2 path (audit B1)

gmhelmold:
## What & why (brutal-audit B1 — the #1 launch blocker)

`R2CasHandler` — the **production** durable CAS path — stored bytes under the client-claimed digest **without computing `blake3(bytes)`**, and served bytes from R2 **without re-verifying** them. So:

- Any client (or a buggy uploader) could `PUT` arbitrary bytes under a fabricated digest → **cache-poison** every subsequent reader of that digest.
- Silent R2 bitrot / storage-tier tampering was served as **trusted** content.

The route doc claimed *"the handler enforces hash equality"* — but only the **in-memory fake** enforced it (`fake_hash`), which is why tests were green while prod was blind. The `HashMismatch` (→422) variant existed and was mapped, but the R2 handler never produced it.

This is the product's core "content-addressable" guarantee being **untrue on the live path**.

## Fix
Added `verify_content_hash` (canonical BLAKE3, constant-time compare, malformed-claim ⇒ mismatch) and wired it into **both**:
- `<R2CasHandler as CasWriteHandler>::write` — verify **BEFORE** the PUT; poisoned bytes are **never persisted**.
- `<R2CasHandler as CasReadHandler>::read` — **re-verify** bytes returned by R2 before serving.

A mismatch emits a `CorrectnessViolation` audit + a `CorrectnessCas` SLI failure and returns `422 HashMismatch`.

**One gate, all surfaces:** every CAS write/read path — native `/v1/cas`, the **Bazel REAPI v2 bridge**, and **sccache** — funnels through `R2CasHandler`, so this single change closes the hole across all of them.

## Tests (5)
`verify_content_hash` accept / mismatch (reports the true actual hash) / malformed-claim; a **write-path regression** proving poisoned bytes are rejected with `HashMismatch` **before any storage I/O**; and an honest-bytes write that passes verification and only then fails at the (unreachable stub) network layer.

## Scope note
Action-Cache (`R2AcHandler`) result-payload integrity is a **separate** follow-up: the AC action-digest is a *key*, not a content hash of the payload, so its integrity needs the `corelink-ac` Merkle/HKDF signed-result envelope wired into the update/lookup path (tracked, not in this PR).

## Verification
Local foreground compile was deliberately **not** run (the founder's shared-Mac runners were saturated by an unrelated mutation-testing storm + sibling projects; compiling the cold worktree would risk the machine). Verification is via the **per-PR CI** corelink-server build/test (the sensible check for this crate; touches no `Cargo.lock` so it does not fire the full per-crate matrix). **Do not merge until CI is green.**


gmhelmold:
### ✅ Local verification PASSED (shared-Mac runners still queue-saturated)

Ran the corelink-server crate compile + tests locally in an isolated worktree (toolchain 1.91.1, `CARGO_BUILD_JOBS≤4`):

- **Full crate:** `cargo test -p corelink-server` → compile clean, **321 passed / 0 failed** (no regression).
- **This change's tests:** `cargo test -p corelink-server --lib storage::r2_s3` → **8 passed / 0 failed / 1 ignored** (the ignored one is the live-R2 round-trip needing creds). Includes:
  - `verify_content_hash_accepts_matching_digest` / `_rejects_mismatch_and_reports_actual` / `_rejects_malformed_claim`
  - `r2_cas_write_rejects_poisoned_bytes_before_storage` — the load-bearing security regression (poisoned write rejected with `HashMismatch` **before any storage I/O**)
  - `r2_cas_write_with_honest_bytes_passes_verification` — confirms honest bytes pass the gate and proceed to storage

Note: the first local attempt hit the known shared-Mac `os error 2` build-race (a dep, `hashbrown`/`aws_sdk_kms`, not this code) under runner contention; a warm-target retry compiled clean. One self-introduced test bug (a `block_in_place` test needed the `multi_thread` runtime flavor — prod is `#[tokio::main]` multi-thread) was fixed in the amended commit.

So this is verified-correct locally; CI just needs a runner slot.

gmhelmold:
### ✅ Load-bearing gates verified LOCALLY (runner pool strangled by non-CoreLink load)

The shared self-hosted Mac runners have been at ~5% effective throughput for 8h+ — individual jobs (`corelink-hash`, `license-policy`) stuck *in_progress* for **6–8 hours** because the box is CPU-starved by non-CoreLink processes (java ~100%, kyverno ~50%, node ~135%, sibling projects). So I ran the load-bearing gates locally (toolchain 1.91.1, isolated worktree):

- ✅ **compile** — `cargo test -p corelink-server` builds clean
- ✅ **tests** — 321 crate tests + 8 `storage::r2_s3` tests pass (0 failed)
- ✅ **`cargo fmt --all -- --check`** — clean (this CAUGHT a real formatting miss in the test code, now fixed + amended — it would have been a red CI)
- ✅ **`cargo clippy -p corelink-server --all-targets -- -D warnings`** — clean, `Finished in 2m07s`, zero warnings

Not run locally (supply-chain / policy gates, near-zero risk for this change — adds **no dependencies**, only a verification function + tests): `cargo-deny`, `cargo-audit`, `cargo-license`, `proptest-density-gate`, INV-registry cross-validation.

Conclusion: the correctness-critical gates (build + test + fmt + clippy) are green locally. The change is ready to merge pending only CI runner capacity.

PR #214, opened 2026-06-10
fix(ci): bump cargo-deny 0.16.4 → 0.19.8 for CVSS 4.0 advisory parsing (repo-wide red gate)

gmhelmold:
## What
Bumps **cargo-deny `0.16.4` → `0.19.8`** (prebuilt via SHA-pinned `taiki-e/install-action@fd2f5e3d…` v2.81.9) across all four call sites: `cargo-deny.yml`, `dependabot-policy.yml`, `cas_foundation.yml`, `lockfile-diff.yml`. Plus ADR-S12-045 amendment (v1.2.0) + CHANGELOG + template doc.

## Why — repo-wide red gate
cargo-deny `0.16.4`'s bundled `rustsec`/`cvss` cannot parse **CVSS 4.0** advisory vectors. The new `RUSTSEC-2026-0073` (libcrux-poly1305) advisory uses `CVSS:4.0/…`, so `0.16.4` hard-fails at advisory-database load (`failed to load advisory database: parse error … TOML parse error`) on **every PR touching `crates/*/src/**`**. The advisories gate has been red repo-wide.

`lockfile-diff.yml` additionally switches from `cargo install` (from-source) to the prebuilt binary, removing the os-error-2 compile race that flakes on the contended self-hosted Mac.

## Security review — §14.s12.004.1 / ADR-S12-045 v1.2.0 (owner-approved 2026-06-10)
- **Strictly positive posture:** `0.19.8` parses newer advisories (incl. CVSS 4.0) that `0.16.4` silently fails to load → catches *more* vulnerabilities, not fewer. No detection regression.
- **No policy change:** verified locally against the unchanged `deny.toml` → `advisories ok, bans ok, licenses ok, sources ok` (only 2 pre-existing `unmatched-skip` warnings).
- **Supply chain:** prebuilt binary via SHA-pinned action; reproducible; no new trust surface.
- Governance recorded in `ADR-S12-045` Amendment 1.2.0 (the "critical fix available" out-of-cadence bump path).

## Verification (local)
- `actionlint` — clean (4 workflows)
- `validate_specs.py` — 465/0
- `validate_canonical_consistency.py` — exit 0, orphans 0
- `cargo-deny check` @ 0.19.8 — `advisories/bans/licenses/sources ok`

## Note
Out of scope (separate concern): the recurring `cargo-audit` / `cargo-mutants` os-error-2 **install races** on the shared Mac (compile-from-source flake, not CVSS). Can follow up by moving those installs to taiki-e prebuilt binaries too.


gmhelmold:
Cross-session note: a parallel session independently built the same fix (#216, now closed as duplicate — this PR is earlier and pins the same 0.19.8). Empirical findings from that work, for this PR's review: (1) 0.16.4, 0.18.2 AND 0.18.3 all fail the CVSS-4.0 parse — 0.19.8 is the first good version, so the pin target here is correct; (2) the SHA-pinned taiki-e/install-action v2.49.49 manifest tops out at cargo-deny 0.18.2, so 0.19.8 installs via the action's cargo-binstall FALLBACK (verified in the pinned main.sh — warn + binstall, no bail). Dependabot #164 bumps the action to v2.81.8 whose manifest carries ≥0.19.8 natively — merge order: this PR first, then #164; (3) `cargo deny check` on the workspace with 0.19.8: advisories ok / bans ok / licenses ok / sources ok — ZERO advisories; the workspace is NOT affected by RUSTSEC-2026-0073 itself; (4) the host ~/.cargo/bin/cargo-deny 0.16.4 (installed by CI, shadows brew's 0.19.8 for `cargo deny`) was replaced with 0.19.8 — each CI run of CURRENT main re-clobbers it with 0.16.4 until this merges.

PR #217, opened 2026-06-10
fix(ops): pin wrangler 4.x at the repo root (stale 3.114.17 broke every prod CLI op)

gmhelmold:
## Problem (Task #44)

Bare `npx wrangler` at the repo root resolved a stale **wrangler 3.114.17** and died parsing the root `wrangler.toml`:

```
✘ [ERROR] Processing wrangler.toml configuration:
    - "containers" should be an object, but got an array
```

(the `[[containers]]` array is v4-only syntax). Every prod CLI op that relied on root-local resolution was broken; the known-good version is **4.95.0**, already pinned in `worker/`.

## Root cause

An out-of-band npm install of `@cloudflare/next-on-pages@1.13.7` (~2026-05-14) dropped **real npm-shaped directories** (`node_modules/wrangler@3.114.17`, `node_modules/@cloudflare/next-on-pages`) into the otherwise pnpm-managed root `node_modules` — `npm ls` marks both **extraneous** (no entry in root package.json), and `next-on-pages` does not exist anywhere in `pnpm-lock.yaml`. Because the root `package.json` declared **no** wrangler devDependency, the stale v3 shim in `node_modules/.bin/` won every bare `npx wrangler` resolution. (Separately/legitimately, `apps/analytics-worker` pins `wrangler ^3.62.0` in its own sandbox — that keeps a wrangler@3.114.17 entry in the lock, scoped to that app only.)

## Fix

- **package.json** — add root devDependency `"wrangler": "^4.95.0"`; `pnpm install` (with `--config.resolution-mode=lowest-direct`) locks it to **4.95.0**, the exact version `worker/` uses (no second 4.x enters the lock). Root `node_modules/.bin/wrangler` is now always 4.x → bare `npx wrangler` is safe everywhere, which also transitively repairs `scripts/e2e-clerk-signup.sh` (the one script using bare `npx wrangler`).
- **scripts/apply-d1-migrations-prod.sh** — `WRANGLER_CMD` default changes from `npx wrangler@latest` to the repo-pinned `$REPO_ROOT/node_modules/.bin/wrangler`; the `WRANGLER` env override is preserved; header exit-code 127 note + error message updated to point at `pnpm install`.
- **scripts/deploy-pages-docs-prod.sh** — checked, **no change needed**: it already resolves `${WRANGLER:-${REPO_ROOT}/node_modules/.bin/wrangler}` via `scripts/_pages-deploy-common.sh:22`, and that helper's "Run: pnpm install (from repo root) to install devDependencies" hint becomes truthful with this PR (previously root had no devDependencies at all).
- **CHANGELOG.md** — `[Unreleased]` → `### Fixed` entry (changelog gate).
- **pnpm-lock.yaml** — root importer gains wrangler; plus two mechanical pnpm-authored refreshes from running a real (non-frozen) install: chromedriver 149.0.1→149.0.2 (transitive of the axe a11y tooling) and eslint peer-suffix canonicalization. No other version semantics changed.

## Verification

| Check | Result |
| --- | --- |
| `node_modules/.bin/wrangler --version` (repo root) | **4.95.0** |
| `npx wrangler d1 migrations list CONFIG_DB --env prod --remote` (no secrets passed) | **config parses under v4**; read-only list succeeded via operator's cached OAuth ("No migrations to apply!") |
| Same command via the stale v3 binary | reproduces the `"containers" should be an object, but got an array` parse error |
| `bash -n` both scripts | clean |
| `bash scripts/apply-d1-migrations-prod.sh` (default dry-run, no remote mutations) | lists 61 migrations, exits clean |
| `python3 scripts/validate_specs.py` | 465 validated, 0 failures |
| `python3 scripts/validate_secrets_matrix.py` | exit 0 |
| `bash scripts/secrets-checklist-verify.sh` | OK (no drift) |

No prod writes were performed (dry-run + read-only list only).

## Follow-ups (out of scope for this card, noted for the backlog)

1. **Main checkout hygiene**: the live checkout's `node_modules` still physically carries the foreign v3 dirs until a one-time `pnpm install` is run there post-merge (pnpm will then own/replace the root wrangler entry + `.bin` shim).
2. **`npx wrangler@latest` sweep**: ~8 other prod scripts (`put-secrets-prod.sh`, `rollback-prod-corelink.sh`, `push-container-prod.sh`, `e-day-container-canary-promote.sh`, `put-secrets-regional.sh`, `verify-secrets-deployed.sh`, `cutover-checklist-prod.sh`, `d-day-secrets-rotate-prod.sh`) use `npx wrangler@latest` — they dodge the v3 landmine (registry-latest) but are unpinned; consider pointing them at the repo-pinned binary in a follow-up. `d-day-migrations-apply-prod.sh` already self-defends with `npx wrangler@4`.
3. **`apps/analytics-worker` is still on wrangler v3** (`^3.62.0`, its own sandbox) — a v4 bump there is a separate change since its config/workflow may need the v4 syntax migration.


gmhelmold:

PR #222, opened 2026-06-10
fix(signup-worker): fail loud on checkout.session.completed missing metadata[tier] (task #39)

gmhelmold:
## What
Closes the last defect from the **task #39 TS-edge sweep**. Two fixes in `apps/signup-worker/src/webhooks/stripe.ts`:

1. **Money path — fail loud on missing `metadata[tier]`.** A paid `checkout.session.completed` whose metadata lacked `tier` was silently defaulted to `"starter"` and persisted to `tenant_billing` — a Max ($149) checkout would be billing-rowed as Starter ($35) with **no operator signal**. Our checkout backend always sets `metadata[tier]` (`corelink-stripe-real client.rs:630`), so a session without it is anomalous: we now return **500 `checkout_missing_tier` before any write**, mirroring the existing missing-tenant/customer fail-loud guard, so Stripe redelivers and the anomaly surfaces in the dashboard.
2. **Honest cancel analytics.** The `subscription_canceled` event hardcoded `from_plan: "starter"` for every cancel; it now derives the real prior plan via the existing `resolveSubscriptionTier` (metadata[tier] → price→tier map → `"unknown"`).

## Sweep context (task #39)
The originally-flagged holes — `quota.ts` missing the `max` tier and stale `pricing.ts` — were **already fixed on main** by the prior sessions (verified: `quota.ts` enumerates the 6-tier ladder + documented legacy `team`/`org`; both `pricing.ts` files carry the canonical $0/$15/$35/$50/$149 rate card). This PR closes the remaining defect the sweep surfaced.

## Tests
- New: `checkout.session.completed` without `metadata[tier]` → 500, **zero** billing/tier writes.
- Extended: `subscription.deleted` asserts `from_plan: "pro"` from the fixture (was unasserted).
- Fixtures that leaned on the buggy default (`metadata.plan` — a key the handler never read) now send `metadata.tier`.
- Local: **76/76** signup-worker tests, `tsc --noEmit` clean.

## Risk
Behavior change only for sessions NOT created by our checkout backend (which always sets `tier`). Those now 500-redeliver visibly instead of silently mispricing — the same posture this handler already takes for missing `tenant_id`/`customer`.


gmhelmold:

PR #225, opened 2026-06-10
fix(signup-worker): payment_status gate + BILLING_DB fail-closed + price↔tier assert + terminal-state guard

gmhelmold:
## Summary

Four audited money-path fail-safety fixes in the Stripe webhook handler (`apps/signup-worker/src/webhooks/stripe.ts`). **Stacks on #222 — merge AFTER #222 lands** (this branch is based on `fix/stripe-webhook-tier-failloud`; once #222 merges, this PR's diff reduces to the four fixes below).

### 1. `payment_status` activation gate (+ async payment events)
`checkout.session.completed` activated entitlement without ever reading `payment_status` — async payment methods (SEPA debit, ACH, …) complete the session with `payment_status='unpaid'` and the money may never arrive. The activation block (tenant_billing upsert + tier_selections activation + `paid_subscription_started` MRR emit) is now one shared helper (`queueCheckoutActivation`) and runs ONLY when `payment_status` is `paid` / `no_payment_required`:
- `'unpaid'` → 200, ZERO writes (entitlement comes later).
- unknown/absent → 500 fail-loud (never grant on an unproven payment, never silently ack an unknown session shape).
- **New:** `checkout.session.async_payment_succeeded` → runs the identical shared activation (this is how delayed payments eventually activate), inheriting all the completed-path validation.
- **New:** `checkout.session.async_payment_failed` → log + 200, no writes (nothing was granted at completed-time, nothing to revoke).

### 2. `BILLING_DB` fail-closed
A missing binding no-op'd every money-path write yet still returned 200 — Stripe never retries a 2xx, so a PAID checkout was silently dropped with no recovery. The handler now mirrors the `STRIPE_WEBHOOK_SECRET` check: 503 `billing_db_unbound` before any processing, so Stripe retries until the deploy misconfiguration is fixed. Inner `if (env.BILLING_DB)` guards retained for TS narrowing. (`wrangler.toml` binds BILLING_DB in all envs — this only bites on a config regression, which is exactly when it must fail loud.)

### 3. price↔tier defense-in-depth assert
`metadata[tier]` is server-set at checkout, but a checkout-backend bug could desync it from the price actually subscribed. Inspected the payloads: the `checkout.session` webhook object carries **no price-shaped data** (`line_items` only exists on API retrieval with `expand[]`; no `items`/`plan`), and we make no Stripe API calls from the webhook — documented in-file. The cross-check therefore lives on `customer.subscription.created`/`.updated`, the first payloads carrying BOTH `metadata[tier]` and the price: when both resolve and disagree → `console.error` + 500 `subscription_tier_price_mismatch` BEFORE any write (redelivery keeps the desync visible; we never pick a winner).

### 4. `tenant_billing` terminal-state guard
Stripe webhooks are unordered — a late `customer.subscription.updated(status=active)` arriving AFTER `customer.subscription.deleted` resurrected `tenant_billing` to `'paid'`. `updateBillingSubscription` (and the `past_due` writer `updateBillingStatus`) now carry `AND status != 'canceled'`, matching the canonical `tier_selections` gate behavior (activation only ever happens via checkout); only a NEW checkout (`upsertBillingPaid` upsert) can move a canceled row forward.

## Tests

- signup-worker suite: **76 → 87 passing** (stripe.test.ts 42 → 53, +11 new): unpaid-completed zero-writes; `no_payment_required` activates; unknown `payment_status` 500; `async_payment_succeeded` activates identically to paid-completed (billing upsert + activation + MRR emit) and inherits the tier fail-loud; `async_payment_failed` no-writes; unbound-DB 503; mismatch 500s on created + updated; agreeing tier+price no-false-positive; deleted-then-late-active resurrection pin; payment_failed terminal guard.
- `npx tsc --noEmit` clean.
- Existing checkout fixtures updated to carry `payment_status: "paid"` (the gate is fail-closed by design); the two signature-rejection tests now bind a DB so they still exercise signature rejection (and additionally pin zero side effects).


gmhelmold:

PR #226, opened 2026-06-10
fix(admin-ui): enforce Clerk auth in middleware + real prod auth provider + ClerkProvider grouping + rate-limit + JWT redaction

gmhelmold:
## What

Five audited auth-enforcement fixes in `apps/admin-ui` — auth was decorative end-to-end: the middleware never enforced, and the production auth provider rejected everyone.

### 1. `middleware.ts` — real enforcement
The `clerkMiddleware` handler ignored its `auth` argument and always returned `next()`. Protected (non-public) paths now call `await auth.protect()` → unauthenticated requests redirect to sign-in. Security headers (per-request nonce + CSP) are applied to **every** response *including Clerk's redirects*; Clerk's response is now re-wrapped (`new NextResponse(result.body, result)`) so its `set-cookie`/handshake headers are preserved instead of dropped (the old code rebuilt a bare `next()` and discarded them). The no-publishable-key graceful fallback (dev/test/E2E) is unchanged.

Turning on enforcement required making the route matcher honest about what is genuinely public — otherwise the launch funnel breaks:
- `/` landing page (its own doc comment already *claimed* it was exempt — it wasn't),
- locale-prefixed `pricing` / `legal` / `privacy` / `security` / `403` (locale-stripping match; `/en/pricingx`-style prefix tricks stay protected),
- `/api/newsletter/subscribe` (deliberately public, docs-site cross-origin POST — now rate-limited, see fix 4),
- new **self-gated** class for `/upgrade`: `clerkMiddleware` runs (server `auth()` must resolve) but no `protect()` — the page owns its `?plan=`-preserving `/sign-in?redirect_url=…` round-trip by documented design.

### 2. `src/lib/auth.ts` — real production auth provider
`defaultProvider` returned `role: null` for everything outside E2E mode, so `RbacGuard`/`CustomerGuard` rejected every real user (dashboards dead). Non-E2E now resolves the real Clerk session via lazy `@clerk/nextjs/server` `auth()`:
- `orgRole` → `ClerkOrgRole` mirroring the E2E mapping (`org:` prefix stripped; explicit `corelink-admin`/`corelink-viewer`/`corelink-member` 1:1; legacy `admin` → `corelink-admin`, `basic_member`/`member` → `corelink-member`),
- solo users with no active org (the self-serve per-user-tenant case) get Viewer-minimum `corelink-member`; **`corelink-admin` only ever comes from an explicit org role** — RbacGuard stays fail-closed,
- `user_id`/`org_id` from the session; `mfa_verified_at` derived from the session-token v2 `fva` claim (`-1`/absent → null → `mfaFresh()` fails closed),
- any error → empty context (fail-closed). The double-gated E2E cookie path and `setAuthProvider` are byte-for-byte preserved.

### 3. Route grouping — ClerkProvider actually mounted
`[locale]/customer/*` and `[locale]/admin/*` lived outside the `(authenticated)` route group, so `ClerkProvider` wasn't mounted and client `useAuth`/`useUser` were dead. Both moved under `(authenticated)` (route groups don't change URLs); one test import path updated; `tsc` clean.

### 4. `/api/newsletter/subscribe` — rate limiting
Public unauthenticated POST had none. Now per-IP via the existing `checkRateLimit` (same limiter + window as `/api/csp-report`), keyed on `CF-Connecting-IP` (edge-set, not client-spoofable) with XFF/X-Real-IP fallback; 429 carries `retry-after` + CORS headers so the docs-site fetch can read it. The deliberate Resend-4xx → 200 masking is untouched.

### 5. `api-client.ts` — JWT redaction
`redactTokens` matched CoreLink PATs but not JWTs; Clerk session tokens echoed in upstream error bodies would leak. Added `eyJ[\w-]+\.[\w-]+\.[\w-]+` scrubbing (`jwt_***_REDACTED`); unit test extended (both shapes in one body; non-JWT `eyJ` text untouched).

## Verification

- `tsc --noEmit`: **clean (0 errors)**
- vitest: **64 files / 335 tests, all green**
- Next/OpenNext production build: **not run locally** (CI covers; shared-Mac load)
- Playwright E2E: **not run locally** (CI covers; E2E mode is key-less so the middleware fallback preserves the fixture flows)

## Dashboard data-layer assessment (no code in this PR)

The dashboards remain honest-but-inert in prod: `AdminClient`/`CustomerClient` are constructed with no `getToken` (no `Authorization` header) and default to `/api/v1/*`, whose catch-all is the E2E mock that returns **503 in production**.

**(a) What exists on the backend today**
- `/v1/customer/*` exists in `crates/corelink-container/src/routes/customer.rs` and matches the `CustomerClient` surface exactly (overview, usage, audit, billing, billing/portal, keys list/create, `keys/:pat_id/revoke`, team, team/invite) — **but** it's backed by `InMemoryCustomerHandler` (non-persistent; "D1-backed handler is a follow-up" per the file) and is reached through the worker's `customer_v1` arm which authenticates by **PAT** (tenant derived from PAT, not from a Clerk session).
- `/v1/admin/audit`, `/v1/admin/ops`, `/v1/admin/tenants` (what `AdminClient` calls) **do not exist** — the container only has generic `/v1/admin/read/:resource` + `/v1/admin/mutate` (`routes/admin.rs`) and `/v1/admin/pilots*` (`routes/admin_pilot.rs`).
- Precedent worth reusing: the worker's `/v1/onboarding/*` arm already verifies **Clerk session JWTs** (`@clerk/backend` `verifyToken` + issuer/azp pinning + D1 `clerk_user_id → tenant_id`, `worker/src/index.ts` ~1229-1390).

**(b) Minimal honest wiring (customer dashboard first)**
1. Frontend: construct `CustomerClient` with `getToken: useAuth().getToken` (works now that ClerkProvider is mounted, this PR) and point `NEXT_PUBLIC_CORELINK_API_URL` at the real API origin.
2. Worker: extend the `customer_v1` arm to accept a Clerk session JWT using the verification + tenant-lookup path already built for onboarding (Authorization: Bearer <clerk-jwt> → tenant_id → inject the same internal headers the PAT path injects). No new auth machinery.
3. Container: replace `InMemoryCustomerHandler` with the D1-backed implementation (the declared follow-up) so overview/usage/keys reflect real tenant state.

**(c) Scope estimate**
- Customer dashboard: ~3-5 days (½ day frontend token plumb; ~1 day worker Clerk-auth reuse on `customer_v1`; the bulk is the D1-backed handler + tests).
- Admin operator dashboard: the backend endpoints it calls don't exist — ~1-2 weeks of backend build (audit query/export, dual-approval ops, tenant directory) before any wiring is honest. Recommend shipping the customer path first and keeping the operator surface explicitly dark.


gmhelmold:

PR #227, opened 2026-06-10
fix(container): fail-closed method gates, users/me tenant parity, brew pre-store integrity

gmhelmold:
## What

Three audited fail-closed findings in the container crate, fixed at the root with tests — no `#[allow]` bypasses.

### 1. Method-gate fail-open default (4 sites)
`routes/cargo.rs`, `npm.rs`, `pip.rs`, `brew.rs` all ended their method-dispatch scope gate with `_ => true`, so an unmapped HTTP method bypassed the per-operation scope check and relied on the adapter's routing to 405 it. Checked each adapter's registrations first: cargo routes GET/PUT/HEAD; npm/pip/brew route GET only (axum `get` also serves HEAD) — no surface legitimately receives OPTIONS (CLI clients only, no CORS), so the blanket-deny is a pure safety net today and a guarantee that a future adapter route can never ship without an explicit scope decision. All four now end `_ => false`, with per-gate tests: DELETE and PATCH carrying a full `cas:rw` scope → 403.

### 2. `/v1/users/me` tenant parity
`routes/users.rs` defaulted a missing/sentinel tenant to an `"_unknown"` echo — the one v1 surface off the fail-closed posture. The handler now takes the shared `AuthTenant` extractor (the same one cas/ac/turbo use): missing or sentinel `x-corelink-tenant-id` → 401 before the body runs. Echo-only correlation metadata (token prefix / route kind) keeps soft defaults. Tests updated (missing → 401, all five sentinels → 401, soft defaults under a real tenant).

### 3. brew `_public` pre-store integrity (was `integrity:"best-effort"`)
The brew adapter stored upstream-fetched bottle bytes into the shared `_public` moat namespace unverified — a MITMed/corrupt upstream response would be persisted once and served cross-tenant until the read path self-heals. Assessment: ghcr.io serves Homebrew bottles as **content-addressed OCI blobs** (`…/blobs/sha256:<64-hex>`; by-digest manifests likewise), so the expected digest is already in the request path — the spec's "brew URLs do not embed the digest" claim predates the ghcr.io OCI layout. `BottleService::fetch` (`corelink-adapter-host/src/brew/bottle.rs` — the one place with both the canonical path and the bytes, upstream of every store consumer) now verifies the fetched bytes against the URL-declared sha256 BEFORE the store:
- mismatch → **refused**: 502, nothing stored, nothing served, plus a `corelink.brew.bottle.integrity_mismatch.v1` audit row (expected/computed digests, `outcome:"refused"`);
- match → cache-fill row carries `integrity:"verified-sha256"`;
- non-content-addressed paths (e.g. manifest-by-tag) keep `integrity:"best-effort"` so the SIEM lane stays distinguishable.

`specs/_proposals/adapters/brew.md` §3 updated to match. No new INV-* ids introduced.

## Verification

- `cargo test -p corelink-server`: **416 passed, 0 failed, 7 ignored** (340 unit + 75 integration + 1 doc-test)
- `cargo test -p corelink-adapter-host`: all suites green (218 unit + brew/cargo/npm/pip/oci integration incl. 2 new adversarial tests: MITM mismatch refused+audited+never stored; honest digest verified+stored)
- `cargo clippy -p corelink-server -p corelink-adapter-host --all-targets -- -D warnings`: clean
- `python3 scripts/validate_specs.py`: green (465 validated); secrets gates: OK, `code_only=0`


gmhelmold:
Landed on main via integration bundle **#238** (bundle/rust-container-2026-06-10, merged 2026-06-11) — one squash commit per source branch, full local verification in the bundle PR comment. Branch deleted.

PR #220, opened 2026-06-10
fix(ci): install cargo-audit/mutants/nextest as prebuilt binaries (kills os-error-2 race)

gmhelmold:
## What
Switches **all 13 from-source `cargo install` sites** for `cargo-audit`, `cargo-mutants`, and `cargo-nextest` to the SHA-pinned **`taiki-e/install-action@fd2f5e3d…` (v2.81.9) prebuilt binary**. Plus ADR-S12-045 amendment (v1.3.0) + CHANGELOG. **Versions unchanged** — pure install-mechanism swap.

## Why
`cargo install <tool>` compiles from source, pulling the heavy `aws-lc-sys` C/asm build. On the contended self-hosted Mac this repeatedly hits a filesystem race (`No such file or directory (os error 2)` / `clang: no input files`), **cancelling/failing the cargo-audit + cargo-mutants gates** — the documented-flake admin merges this session (#206, #208, #214) were all forced by it. Prebuilt binaries install in seconds and don't compile, so the race is gone and the gates get lighter.

## Sites (13)
- **cargo-audit** (2): PR gate + daily cron
- **cargo-mutants** (9): per-PR gate, workspace nightly, mutation-nightly, + 6 per-crate nightly
- **cargo-nextest** (1): the PR mutation gate's same step that raced

## Mac safety
The 6 per-crate cargo-mutants jobs are `if: github.event_name == 'schedule'` (nightly only) — **this PR does not run mutation**. Editing those files triggers their normal build/test lanes (which run on any change anyway), not the heavy mutation.

## Governance — ADR-S12-045 v1.3.0 / §14.s12.004.1 (owner-approved 2026-06-10)
cargo-audit is high-risk-lane tooling. **No version change → no detection change.** Supply-chain trust is *narrowed*: a pinned upstream release artifact via the already-trusted taiki-e action+SHA (same one used for cargo-deny) vs. compiling arbitrary crates.io source (incl. `aws-lc-sys` build scripts) at install time. A gate that flakes-red gets bypassed; a reliable gate gets enforced.

## Verification (local)
- `actionlint` — clean (10 workflows)
- `validate_specs.py` — 465/0
- `validate_canonical_consistency.py` — exit 0, orphans 0
- taiki-e v2.81.9 manifest confirmed to carry every pinned version (cargo-audit 0.22.1, cargo-mutants 25.0.1/27.0.0, cargo-nextest)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #252, opened 2026-06-11
fix(billing): close paid-tier-without-payment enforcement hole (CRITICAL)

gmhelmold:
## CRITICAL — paid tier served without payment (e2e adversarial audit 2026-06-11)

A brutal adversarial e2e audit (tracing the full user journey, not surface-by-surface) found a **launch-blocking** hole the prior pentest (#249, "SHIP") missed.

### The live exploit
- The checkout backend persists the **requested paid tier** into `tier_selections` at checkout-**START** with `subscription_state = 'pending_checkout'` — *before any payment* (the container comment at `apps/signup-worker/src/webhooks/stripe.ts` confirms: *"The container persists the row as `pending_checkout` when checkout starts"*).
- The **canonical access gate** the whole system assumes is `subscription_state = 'active'` (the live Stripe handler only flips the row to `active` on a **paid** `checkout.session.completed`).
- But the request-time tier-enforcement read — `worker/src/lib/quota.ts::getTierForTenant`, wired at `worker/src/index.ts:1597` → `checkStorageQuota`/`checkRequestQuota` — read `tier_selections.tier` with **no state filter**.

**Result:** a user could select a paid tier, **abandon Stripe checkout, and be served full paid quota for free.**

### The fix (minimal, surgical)
- `getTierForTenant` now honours the canonical gate: `... AND subscription_state = 'active'`, falling through to `tenant.tier` → `free` otherwise.
- The container customer-dashboard plan read (`crates/corelink-container/src/customer_d1.rs`) mirrors the same filter so a pending checkout never *displays* as the active plan.

### Cold-check correction (why this PR is small)
The first pass also "fixed" the container's Stripe materializer (price-id fail-loud + past_due downgrade + process-then-claim). A cold-check of the **live** path reverted all of that, because:
- The **live** Stripe tier-writer is the **signup-worker** (`corelink-signup.humangr.com/webhooks/stripe`), which **already** maps real `STRIPE_PRICE_ID_*` ids, downgrades on `past_due` → `inactive`, and is process-then-claim idempotent. No webhook-handler change is needed.
- The container's `/v1/billing/stripe-webhook` is **secondary** (duplicate Stripe endpoint → removed under task #46). Activating it would **split-brain** `tier_selections` writes with the canonical handler, and the price-id fail-loud risked **bricking the entire container startup** over a secondary billing config.

So the single live bug was the enforcement read — and that is all this PR touches.

## Verification
- `worker/tests/quota.test.ts`: **20/20** (+2: `pending_checkout` → `free`, `active` → paid). `tsc --noEmit` clean.
- `customer_d1.rs`: SQL-string-literal change; the existing `billing_happy_path` test passes (its mock matches the query by substring, independent of the `WHERE` clause). Container compile confirmed via CI on this PR (local single-crate verify was hampered by the documented shared-Mac `os error 2` concurrent-target race — PR #220 territory; CI's runner workspace is the canonical gate).
- Auth / tenant-isolation / PAT-revocation re-audited clean; **unchanged**.


gmhelmold:
**Admin-merge — documented infra reason ([local-verify-when-CI-starved], owner-authorized 2026-06-05).**

`pre-merge-gate-check.sh 252`: **0 failing, 4 pending** (`cargo-audit`, `cargo-deny`, `cargo-mutants`, INV cross-validate) — all **starved** on the saturated shared self-hosted Mac runners (queued, not failing; cargo-audit also hit a 10-min cancel-flake on the first run).

Why the 4 pending gates are deterministically green / non-blocking for this PR:
- **Dep-free change** — diff vs main is exactly 4 files (CHANGELOG, customer_d1.rs, quota.ts, quota.test.ts); **Cargo.lock is byte-identical to main** (verified `git diff main…HEAD -- Cargo.lock` = empty). So `cargo-audit` + `cargo-deny` scan the same lock as green `main` → identical result.
- **No INV annotations touched** → INV cross-validate is structurally green.
- **cargo-mutants** changed-lines surface is a single SQL **string literal** + a comment in customer_d1.rs → negligible mutable surface; the security behaviour itself is covered by `worker/tests/quota.test.ts` (20/20).
- The load-bearing security fix (`quota.ts::getTierForTenant`) is **fully verified locally**: vitest 20/20 (+2 for the active-state filter) + `tsc` clean.

Launch-blocking CRITICAL (paid tier served without payment) — landing per owner's anti-CI-orbiting guidance.

PR #264, opened 2026-06-13
fix(security): CAA-360 full remediation — 35 findings fail-closed + M2 max_concurrency

gmhelmold:
## CAA-360 adversarial audit → full remediation (LIVE in prod 2026-06-13)

A 16-agent 360° audit (3 Opus pentest + 3 Opus + 5 Sonnet + 5 Haiku, adversarially verified —
12 false-positives refuted) found a coherent root-cause class: **secrets/invariants that fail
OPEN and silent.** This PR remediates **all 35 confirmed findings** (every severity) to fail
**CLOSED and loud** across 26 code files, decomposed into 11 disjoint-file WPs.

### Highlights
- **F1/F2** R2 tenant-prefix TDK mandatory on prod path; public-UUID raw fallback removed
  (cfg(test) only) — closes same-ms cross-tenant CAS/AC co-residence.
- **F5** R2AcHandler re-enforces divergent-body invariant (no silent AC poisoning).
- **F7/F8** CAS residency-aware (R2_CAS_REGION) + worker→container env forward-list completed.
- **F9/F18/F28/F15** ERASURE_SALT_KEY / PAT_SIGNING_KEY / internal-auth fail-closed + ≥32 floors.
- **F6/F10** Stripe success_url/cancel_url host-allowlist (open-redirect).
- **F22** timingSafeEqual real per-isolate HMAC key, no length branch. **F37** Vault Debug redaction.
- quota/session fail-closed, OCI/Turbo bounds, brew validation, doc-vs-code reconciliations.
- **M2** introspect `max_concurrency` (ratified byte-compatible shape with corelink-runners).

### Verification (all green, local)
`cargo check` · `clippy -D warnings` · worker+signup `tsc` · `cargo test -p corelink-server`
· `validate_specs` 474/0 · migrations additive.

### Deployed
Container `4b9e9ccf-r1` (digest 045f926e) ×5 + signup-worker. Secrets rotated to ≥32
(internal-auth + PAT-signing) + R2_TDK_HEX/ERASURE provisioned. Smoke: health 200 ×5,
money-path/DSR/cache mounted + auth-gated (≥32 floors pass).

Tracked architectural/infra follow-ups: `docs/security/2026-06-13-CAA-360-followups.md`.
Full chewed report: `docs/security/2026-06-13-CAA-360-audit-report.md`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #266, opened 2026-06-14
fix(security): CAA-360 wave-1 follow-ups — write-gate can_write, fail-closed storage, adapter hardening

gmhelmold:
The next CAA-360 remediation tranche on top of the 35-finding pass (#264):

- **F27** — cargo/brew/npm/pip cache **write gate** now requires the PAT's `can_write` capability, derived from a SINGLE PAT verification via the resolver port (`resolve_with_capability`) rather than a redundant second verifier, in addition to the server-trusted `x-corelink-scope` header (two layers, one verification).
- **F8** — OCI adapter rejects session-table exhaustion with `429 Too Many Requests` + `Retry-After` (`TooManyOpenSessions`).
- Storage handlers that fail to construct resolve to a loud `UnavailableCas/AcHandler` → **503** (fail-closed; never silent 500/empty-200).
- New `scripts/check-env-contract.py` CI gate greps every container `env::var` against the Worker DO forward-list.

Verified locally pre-cherry-pick: `cargo check --tests` + `clippy --all-targets -D warnings` + `cargo test` (481 + adapter suites) all green. Wave-2 design pinned in `docs/security/2026-06-13-CAA-360-wave2-design.md`.


gmhelmold:
Superseded by #268 — its two commits (ca8aba63 + 90b20dd0) are included verbatim in the bundle. Re-audited adversarially there: safe to merge, zero blockers.

PR #317, opened 2026-06-17
fix(auth): wire CAS-erase + DSR to the dedicated CORELINK_ERASE_AUTH_KEY (rt-nuclear #18-21)

gmhelmold:
Nuclear cycle-2 recovery (2026-06-17) confirmed — live on `main` e95cff3c — that the **#297 per-consumer-key split never reached the destructive surfaces**:

- `/_internal/cas/*/erase` was seeded from `internal_auth_key_from_env` (the **ADMIN** key) → an admin-key leak could drive irreversible erases (#20/#21).
- `/_internal/dsr/*` read `CORELINK_INTERNAL_AUTH_KEY` **directly** → the GDPR mass-erase surface honored **no** dedicated erase key (#18/#19).

## Fix
Swap both to `erase_auth_key_from_env()` (`CORELINK_ERASE_AUTH_KEY`) — exactly the LEAD wiring TODO documented at `admin.rs:448-455`. **Non-breaking**: still falls back to the shared `CORELINK_INTERNAL_AUTH_KEY` when the dedicated key is unset (so erase/DSR keep working until secret #158 is provisioned); the **≥32-char fail-CLOSED floor** is preserved on the DSR path.

This makes the #297 split **real on the erase plane**.

## Not closed here (tracked follow-ups, see `docs/security/2026-06-17-redteam-nuclear-cycle2-recovery.md`)
- Remove the shared fallback entirely for destructive consumers — **owner-gated** (needs secrets #156/#157/#158 provisioned first, else erase/DSR unmount).
- Per-tenant authz on erase/DSR (require a D1 `dsr_requested` row; never a body-asserted `tenant_id`).
- The `/_internal/*` CF Service Binding (off the public internet) — owner/infra.

## Verify (local, quiet-ish Mac, `CARGO_BUILD_JOBS=2`, disk freed)
- `cargo clippy -p corelink-server --all-targets -D warnings` → **clean** (exit 0)
- `cargo test -p corelink-server` → **12 suites, 0 failed** (exit 0)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #318, opened 2026-06-17
fix(quota): OCI $-ceiling gate resolves tenant from the verified bearer (rt-nuclear #2/#8/#9)

gmhelmold:
Nuclear cycle-2 recovery confirmed (5/5 votes) the **OCI push surface bypasses the per-tenant monthly $-ceiling entirely**.

## Bug
The Worker strips `x-corelink-tenant-id` on the OCI pass-through (`index.ts:1716-1752`, returns before the quota block). In the container, `oci_quota_gate` keyed the charge on that header behind `!tenant.is_empty()` — **always empty ⇒ the charge was always skipped**. Any Free-tier PAT with OCI push → unbounded billable manifest/blob-upload/finalize work at a flat fee. ADR-0068 cost tripwire inert on OCI; the ~80% margin attacked directly.

## Fix
`oci_quota_gate` now recovers the cost-attribution tenant by **HMAC-verifying the realm bearer** (`corelink_adapter_host::oci::auth::verify` — the same token the data plane checks), threaded via a new `OciCostGate` state carrying a copy of the realm signing key.
- **Verify, not parse** — the bearer's tenant segment is attacker-controlled; an unverified read would let one tenant bill another.
- A write with **no valid bearer is left uncharged** — the data plane 401s it, so no billable work succeeds.

## Test
`oci_bearer_tenant_resolves_only_a_verified_bearer`: valid bearer → its tenant; forged (wrong-key) bearer → None; absent header → None.

## Scope / tracked OCI follow-ups
This PR closes the **$-ceiling tenant-resolution** (the core bypass, #2/#9 + the $ part of #8). Same-cluster follow-ups: per-tenant OCI byte ceiling (#3/#12), `/token` Argon2id cost cap (#4), Worker-edge request-count for OCI (#8). See `docs/security/2026-06-17-redteam-nuclear-cycle2-recovery.md`.

## Verify (local, freed disk)
- `cargo clippy -p corelink-server --all-targets -D warnings` → **clean** (exit 0)
- `cargo test -p corelink-server routes::oci` → **13 passed / 0 failed**
- full `cargo test -p corelink-server` → (running; will confirm before merge)


gmhelmold:
Merging on local-green + documented infra reason (owner-authorized solo execution; same posture as #317). CI: **0 failing**, 10 queued + 1 in-progress behind the saturated shared-Mac runners (load-starved, not validating). Local verify on the freed-disk Mac is authoritative: `clippy -p corelink-server --all-targets -D` clean + full `cargo test -p corelink-server` = **12 suites / 0 failed** (+ the new `oci_bearer_tenant_resolves_only_a_verified_bearer` unit test). Container-only change; non-breaking (a write with no valid bearer is left uncharged exactly as before — the data plane 401s it).

PR #319, opened 2026-06-17
fix(auth): unify self-serve scope classifier — close read-only→read-write self-escalation (rt-nuclear #15)

gmhelmold:
Nuclear cycle-2 recovery confirmed (5/5 votes, **authed-tenant** reachability) that a **read-only PAT can self-escalate to read-write**.

## Bug — divergent scope matching
- Gate `routes::customer::mint_requests_write` → **exact-token** match.
- Persister `customer_d1::map_requested_scopes` → **substring** match (`s.contains("write")`).

A scope like `"writes"` / `"cache:write-x"` is FALSE for the gate (so the read-only caller's scope is never checked) but TRUE for the persister (stores a real `read-write` PAT). A read-only holder POSTs `/v1/customer/keys {"scopes":["writes"]}` → gets a working read-write PAT → writes/poisons/deletes the tenant's CAS, surviving rotation of the original token. Renders the read-only-key feature security-meaningless.

## Fix — one classifier, fail-closed
Both the gate and the persister now derive from a **single exact-token classifier** `scope::classify_requested_scopes` (the source of truth):
- Gate fires for anything **not provably `ReadOnly`**.
- Persister maps `ReadOnly`/`ReadWrite`, rejects `Admin` (as before) **and unrecognized tokens (fail-CLOSED)** — no more substring-defaulting an unknown token to a privilege.

By construction, "gate fires" ⟺ "mint would persist write/admin".

## Test
`classify_requested_scopes_is_exact_token_and_fail_closed`: canonical inputs classify right; `"writes"`/`"cache:write-x"`/`"my-write"`/`"rewrite"`/`"writable"` all **rejected** (never silently `ReadWrite`); read+unknown mix still fails closed.

## Verify (local, freed disk)
- `cargo clippy -p corelink-server --all-targets -D warnings` → **clean**
- `cargo test -p corelink-server scope::` → 17/0
- full `cargo test -p corelink-server` → (running; confirm before merge)


gmhelmold:
Merging on local-green + documented infra reason (owner-authorized solo execution; same posture as #317/#318). CI: **0 failing**, 11 queued behind the saturated shared-Mac runners (load-starved, not validating). Local verify on the freed-disk Mac is authoritative: `clippy -p corelink-server --all-targets -D` clean + full `cargo test -p corelink-server` = **0 failed** (incl. the 3 scope tests: classifier exact-token/fail-closed, gate "viewer"→non-write, persister "writes"→rejected). Container-only change; the escalation is closed at the persister (unknown scope ⇒ 4xx, never silently read-write).

PR #320, opened 2026-06-17
fix(npm): bind cached metadata identity to the requested key — close _public cache-poisoning (rt-nuclear #7)

gmhelmold:
Nuclear cycle-2 recovery confirmed (5/5, **any-free-tenant**) cross-tenant **npm metadata cache-poisoning** of the shared `_public` namespace.

## Bug
Unscoped npm metadata is cached in the SHARED `_public` namespace keyed by the **normalized** name (`trim().to_ascii_lowercase()`), but `refresh_from_upstream` fetches the **raw** path name and never checks the fetched JSON's canonical `name`. A read-only PAT (any tenant) `GET /npm/<t>/<RawName>` where `<RawName>` normalizes to a popular package's `_public` key but resolves upstream to different-identity content → poisons every tenant's view of that package for the TTL (wrong versions / `dist.shasum` → broken/pinned installs). Integrity + availability attack on the moat.

## Fix
`require_metadata_name_matches`: require `normalise(fetched.name) == normalise(pkg)` before the KV write — **fail-CLOSED** with a new `NpmAdapterError::MetadataNameMismatch` (502, non-leaking; mirrors the tarball `IntegrityMismatch` contract). The cache-HIT path **self-heals** — a name-mismatched entry falls through to a fresh fetch rather than being served. Binds the registry's own canonical identity to the KV key, collapsing case/trim/encoding aliasing.

**Not RCE:** tarball bytes are per-tenant namespaced + SHA-verified vs `dist.shasum` before store, so poisoned metadata can't serve malicious bytes.

## Test
`require_metadata_name_matches_binds_identity_to_key`: exact/case/trim → Ok; different-identity `name` → `MetadataNameMismatch` (502); missing/blank `name` → fail-CLOSED.

## Verify (local)
- `cargo clippy -p corelink-adapter-host` + `-p corelink-server --all-targets -D` → **clean**
- `cargo test -p corelink-adapter-host metadata` → 8/0
- full `cargo test -p corelink-server` → (running; confirm before merge)


gmhelmold:
Merging on local-green + documented infra reason (owner-authorized solo execution; same posture as #317/#318/#319). CI: **0 failing** (queue-starved behind the saturated shared-Mac runners). Local verify authoritative: clippy -p corelink-adapter-host + -p corelink-server --all-targets -D = clean; corelink-adapter-host metadata tests 8/0 (incl. require_metadata_name_matches_binds_identity_to_key); full `cargo test -p corelink-server` = **0 failed**. Additive (new fail-CLOSED error variant); container compiles green (no exhaustive-match break).

PR #323, opened 2026-06-17
fix: rt-nuclear batch2 — DSR authz + OCI request-count + Worker quota/erase-key (4 findings, 1 PR)

gmhelmold:
**Bundled** (one PR per owner request — minimize shared-Mac CI triggers). Four remaining nuclear findings across Rust container + privacy-erasure-worker + Worker TS; implemented via 3 parallel file-disjoint agents, integrated here, serialized verify.

| # | Fix | plane |
|---|---|---|
| **#18/#19** | `/_internal/dsr/erase` now binds to a **D1 `dsr_requested` legitimacy pre-check** before any fan-out — a leaked internal key can no longer erase an arbitrary tenant (body-asserted tenant_id). Fail-CLOSED on store error (irreversible op ⇒ ambiguous = deny). | container + privacy-erasure-worker |
| **#8** | OCI writes now metered against the **monthly request-count quota** (container-side `RequestCountGate`, keyed on the same verified bearer tenant #318 uses) — closes the request-count half of the OCI quota bypass. | container |
| **#24** | Worker quota pipeline **reordered** — the cheap monthly request-count check runs first + 429s an over-cap free tenant BEFORE the expensive `getTierForTenant`/storage-SUM D1 reads (cost-amplification). | Worker |
| **#23** | **Real latent mismatch fixed** (not moot): the signup-worker injected the shared key, but the main Worker's `/_internal/dsr/*` gate expects the erase consumer key in the full-split config → both erase surfaces would 401 once a dedicated erase key is provisioned. Added `resolveEraseAuthKey` (erase-first, shared-fallback) mirroring the container. | Worker |

## Verify (local, serialized)
- `cargo clippy -p corelink-server -p corelink-privacy-erasure-worker --all-targets -D warnings` → **clean** (fixed 3 lints the disk-blocked #8 agent couldn't catch)
- `cargo test -p corelink-server -p corelink-privacy-erasure-worker` → **0 failed** (corrected a #8 test's fixed-clock constant T0 that was mislabeled 2026 but was actually 2025 — the year_month/Hinnant impl was correct)
- `worker` tsc + `apps/signup-worker` tsc → **0** (34 + 23 vitest green per the agents)

Cold-reviewed the two security-critical diffs (the DSR `DsrLegitimacyStore` pre-check runs BEFORE fan-out + fail-closes on error; the OCI request-count keys on the verified bearer, not a forgeable header).

## Note (contained, follow-up)
The DSR reject path emits no CloudEvent (no `Rejected` variant in the erasure taxonomy) — the **erase is blocked** (the security property), but auditing the rejected *attempt* is a small additive follow-up (a new `ErasureCloudEventType::Rejected`).


gmhelmold:
Merging on local-green + documented infra reason (owner-authorized solo execution; bundled 4 findings in ONE PR to minimize shared-Mac CI triggers). CI queue-starved behind the saturated runners (not validating). Local verify authoritative: clippy -p corelink-server -p corelink-privacy-erasure-worker --all-targets -D = clean; cargo test (both crates) = **0 failed** (617 container + privacy-worker); worker + signup-worker tsc = 0. The 2 transient #8 test failures were a mislabeled fixed-clock constant (T0 said 2026 but was 2025) — impl correct, constant fixed; re-verified 10/0. Cold-reviewed the DSR pre-check (runs before fan-out, fail-closed on error) + OCI request-count (keys on verified bearer).

PR #426, opened 2026-06-21
fix(brew): send OCI manifest Accept header to ghcr (completes the brew 502 fix)

gmhelmold:
Prior fix (#423) added the ghcr token dance but brew STILL 502'd. Root-caused live: a manifest GET with a valid token but no `Accept` → ghcr 404 → Upstream → 502. Add the OCI/Docker manifest+index media types (+ `*/*`). Verified against ghcr: token-only→404, token+Accept→200. clippy -D clean, 40/40 brew tests.


gmhelmold:
Merging via `--admin`: Linux gates infra-down (3-5s startup fails). Local-verified: clippy -p corelink-adapter-host -D warnings clean + 40/40 brew tests; the token+Accept dance verified live against ghcr.io (token-only→404, token+Accept→200). Per ratified CI=local policy.

PR #430, opened 2026-06-21
fix(brew): make _public dedup namespace storable (final brew-502 causes 4+5)

gmhelmold:
The brew 502 had 5 compounding causes. #1-3 (ghcr token/Accept/path) fixed the upstream fetch; this lands the last two, both in the `_public` cross-tenant dedup namespace: (4) no `tenant_storage_state` row → byte-accounting fail-closed (migration 0073 seeds it per region); (5) `_public` non-UUID → R2 prefix derivation fail-closed → reserved sentinel-UUID prefix (scoped to `_public`, real tenants unaffected). Verified LIVE: brew manifest → 200. New _public-prefix security test; migrations additive; clippy -D clean.


gmhelmold:
Merging via `--admin`: Linux gates infra-down (3-5s startup fails). Local-verified: `cargo test -p corelink-server --lib` incl. the new `public_namespace_resolves_a_stable_isolated_derived_prefix` test (707 pass), clippy -D clean, migrations additive (79 scanned), CHANGELOG updated. The fix is verified working LIVE (brew→200 on diag3). Per ratified CI=local policy.

PR #434, opened 2026-06-21
fix(cargo): route sccache through the 2-level MoatCache (fix HashMismatch 502)

gmhelmold:
cargo/sccache 502'd on every PUT — it passed the sccache key (blake3 of compile inputs, not the output) as the CAS digest, which the integrity-verifying CAS write rejects. Now routes through the same MoatCache brew/npm/pip use, namespaced per-tenant (private). Gains dedup. clippy clean; lib tests compile (708). bazel REAPI v2 + OCI validated working in the same sweep.


gmhelmold:
Merging via `--admin`: Linux gates infra-down. Local-verified: `clippy -p corelink-server` clean, lib test target compiles (708 tests). Will verify the round-trip LIVE post-deploy (container rebuild). Per ratified CI=local policy.

PR #438, opened 2026-06-21
fix(adapters): make cargo/sccache + OCI/docker usable by real clients

gmhelmold:
Real-user-path e2e (signup->PAT->client) found 2 blockers curl missed: cargo normalize_key 400'd sccache's .sccache_check probe (now accepts safe keys; bytes are content-addressed via MoatCache); OCI realm host corelink-oci.humangr.com was never provisioned (added CNAME + Worker route). 9/9 translate tests; clippy clean.


gmhelmold:
Merging via `--admin`: Linux gates infra-down. Local: 9/9 cargo translate tests, clippy clean. DNS+route already applied to prod; container rebuild (cargo key fix) follows. Per CI=local policy.

PR #599, opened 2026-07-03
fix(billing): single source of truth for signup-worker Stripe events + reconcile tool

gmhelmold:
## Problem

The **signup-worker** (`corelink-signup.humangr.com/webhooks/stripe`) is the **authoritative billing + downgrade handler**. But its live Stripe endpoint's `enabled_events` was **hand-configured**, with nothing tying it to the code's dispatch allowlist (`HANDLED_EVENT_TYPES`). Hand-clicked config drifts.

**As-configured live** (endpoint `we_1Tolig…`, per the coordinator's F1 handoff — 5 events) it is **missing `checkout.session.async_payment_succeeded`** → a delayed-payment (SEPA/ACH) checkout that completes `unpaid` and grants on the async event would never provision the paying customer. (The downgrade trio `customer.subscription.{updated,deleted}` + `invoice.payment_failed` IS wired — so this is a grant-gap, not the catastrophic downgrade-loss. But the **drift class is unbounded**: a future accidental drop of a downgrade event silently strands a canceled customer on a paid tier.)

## Root-cause fix (kills the drift class)

- **Single source of truth:** `apps/signup-worker/src/webhooks/handled-stripe-events.json`. `stripe.ts` builds `HANDLED_EVENT_TYPES` from it (**proven to bundle into the worker** via `wrangler deploy --dry-run` — the 7 events land in `index.js`); the reconcile script sets the **live endpoint's** `enabled_events` from the **same file**. Runtime and dashboard can no longer diverge.
- **`scripts/ops/stripe-reconcile-webhook-events.sh`** — idempotent, **dry-run by default**, never reads/hard-codes a key (auth is the Stripe CLI's own — same safety pattern as `stripe-setup-tiers.sh`), prints the account identity for eyeball, lists + **flags duplicate / stray-CLI endpoints** (report-only, never deletes), and re-reads to verify after applying. The **operator** runs the `--yes --live` one-shot with a restricted live key.
- **Guardrail** (`tests/handled-stripe-events.test.ts`) — fails CI if any downgrade-critical event is dropped.
- **Runbook** — `docs/operator/stripe-webhook-events-reconcile.md`.

## Boundary

The admin lane does **not** run this against LIVE — it writes/vets the tool; the operator/coordinator runs the live reconcile (OOB, restricted `rk_live_` key). No secrets touched.

## Verification

- `tsc --noEmit` clean; guardrail **5/5**; wrangler build bundles the JSON-imported events into the worker.
- Full signup-worker suite unchanged — the one pre-existing `dsr_verify_cron` red (`msFromIso` stale assertion) is **identical on clean `origin/main`** (`8f57a160`); it's fixed + the suite gets a CI gate in a follow-up branch.
- Reconcile diff logic **proven against a mocked endpoint list**: correctly surfaces the missing events (`ADD customer.subscription.deleted`, `ADD invoice.payment_failed` in the test scenario) and the stray-listener `WARN`.


gmhelmold:
Merging via `--merge --admin` (merge-commit, NOT squash — the OKF concept's `checkpoint_sha` references commit `25d5a389` which must stay reachable in main history).

**Documented infra reason:** all 7 code-validating gates are GREEN (dco, okf-wiki-validation, changelog, gitleaks, trivy, cargo-deny, cargo-audit). The 4 remaining pending checks are non-validating for this diff: `Apply path-based labels` + `Compute size bucket` run on the congested `[self-hosted, mac]` runner (cosmetic labels only), `dependabot-policy sentinel` is N/A to a non-dependabot PR, and `cargo-license` is a no-op (this PR has zero Rust/Cargo changes — only TS/JSON/bash/docs). Branch protection required-checks = [].

PR #600, opened 2026-07-03
ci(signup-worker): gate the vitest suite in PR CI (+ fix pre-existing stale assertion)

gmhelmold:
## Why

The **signup-worker** is the authoritative billing + downgrade handler, but its vitest suite ran in **no PR workflow** (only the e2e jobs installed it, none ran the unit/integration suite). That left the new **`handled-stripe-events` single-source-of-truth guardrail** (merged in #599 — fails CI if a downgrade-critical Stripe event is dropped) **unenforced**. An unenforced guardrail is theater; this makes it load-bearing.

## What

- **`.github/workflows/signup-worker-vitest.yml`** — ubuntu, `npm install` + `npx vitest run` on `apps/signup-worker/**`, mirroring the `worker-vitest.yml` pattern (SHA-pinned actions reused from it).
- **Prerequisite fix:** the suite had ONE pre-existing red — `dsr_verify_cron`'s `"msFromIso returns 0 on garbage"` asserted `0`, but the impl **intentionally returns `null`** (the caller skips the row on `null`; `0` would be epoch-1970 → a recurring false `sla_breached` page every cron tick). Fixed the stale assertion + its name to match the deliberate impl. This red is **identical on clean `origin/main`** — unrelated to #599; fixing it is the prerequisite to gating a green suite.

## Verification

Full signup-worker suite **182/182 green** locally with the fix, so the new gate lands green.


gmhelmold:
Merging via `--admin`. **Documented infra reason:** all code-validating gates GREEN (vitest [the new self-proving gate], okf, dco, actionlint, Verify-actions-SHA-pinned, secrets-matrix, gitleaks, trivy, cargo-deny, cargo-audit, changelog). The 5 pending are non-validating for this diff: `Apply path-based labels` + `Compute size bucket` (cosmetic, on the congested self-hosted-Mac runner), `Post welcome comment`, `dependabot-policy sentinel` (N/A — not a dependabot PR), and `cargo-license` (no-op — zero Rust/Cargo changes). Branch protection required-checks = [].

PR #601, opened 2026-07-03
chore(billing): source-of-truth + guardrail for the container webhook event set

gmhelmold:
## Why

Closes the container-side of the Stripe webhook `enabled_events` hygiene (owner: "no loose end"). The **container** (grant-only) materializer's live endpoint (`we_1Tfh8P…`, "Corelink prd") is over-subscribed to a **~236-event catch-all**, but it only materializes the **10** events in `EVENT_MATERIALIZATION_MATRIX`. Over-subscription fails safe (extra events are no-op acks) — but it's untidy, and there was no source-of-truth for the container's event set the reconcile tool could use.

## What

- **`crates/corelink-billing-stripe-materializer/container-webhook-events.json`** — the 10 events (same schema the reconcile script consumes: `endpoint_url` + `enabled_events`).
- **Rust guardrail** (`handler.rs::container_webhook_events_json_matches_the_matrix`) — asserts the JSON's `enabled_events` == the `EVENT_MATERIALIZATION_MATRIX` event-type column exactly. `EVENT_MATERIALIZATION_MATRIX` stays the **sole authority**; the JSON is a guarded mirror that can't drift (add an 11th event to the matrix → the test fails until the JSON is updated).
- **Runbook** updated — the reconcile script already supports `--url`/`--json`, so the operator tightens the live endpoint **236 → 10** with the same tool (dry-run first; STOP on any `+ ADD`).

## Safety

**No runtime change** — the matrix is untouched. Tightening 236→10 is safe: events outside the matrix are no-op acks in the container, so narrowing removes only events it never acted on. Under-subscription (the dangerous direction) is exactly what the guardrail prevents from drifting. The container is grant-only; authoritative downgrade stays in the signup-worker.

## Verification

`container_webhook_events_json_matches_the_matrix` passes; `cargo clippy -p corelink-billing-stripe-materializer --tests -- -D warnings` clean.


gmhelmold:
Merging via `--admin`. **Documented reason:** all code-validating gates GREEN — including the Rust ones (cargo-mutants on changed lines, proptest-density, cargo-deny, cargo-audit, INV×TLA×code×tests) + okf + dco + gitleaks + trivy. The 5 pending are non-validating for this diff: labels/size-bucket/welcome-comment (cosmetic, congested self-hosted-Mac runner), dependabot-policy sentinel (N/A), and cargo-license (no-op — this PR adds zero new dependencies). Branch protection required-checks = [].

PR #602, opened 2026-07-03
fix(auth): enforce team_member RBAC role → scope (CAA-360 HIGH)

gmhelmold:
## CAA-360 HIGH — RBAC roles were dead

`verifyClerkSessionAndResolveTenant` never queried `team_member.role`, and the `customer_v1` forward hardcoded `x-corelink-scope: 'read-write'`. So a **`viewer` seat had the same write access as the owner** — the four RBAC roles (0074: owner/admin/member/viewer) were schema-enforced but never consulted. The session/token-exchange mint paths also handed viewers a **write-capable PAT**.

## Root-cause fix (all scope-deriving consumers)

- **`clerk_auth.ts`** — resolver returns `role`: owner path → `owner`; team_member path → the 0074 role, **failing safe to least-privilege `viewer`** on a missing/unknown role; githugr federated login → `owner` (unchanged, no regression). `ClerkAuthResult` gains `role`.
- **`index.ts` `customer_v1`** — derives `x-corelink-scope` from role (`viewer` → `read-only`, else `read-write`).
- **`session_exchange.ts`** — both mint flows (token-exchange + session-exchange) cap a `viewer` **down** to `read-only` regardless of requested scope.

`read-only` is a real container-enforced scope (`scope.rs`): CAS/AC GET allowed, write denied.

## Tests

Worker suite **393/393** (+3 new: `viewer→read-only`, `admin→read-write`, `missing-role→read-only` fail-safe); `tsc` clean.

Part of the pre-HN-launch CAA-360 remediation (`reports/audits/2026-07-03-caa360-pre-hn-launch.md`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #606, opened 2026-07-04
fix(dsr): fan GDPR erase across residency regions; fail-closed completeness (CAA-360 CRITICAL)

gmhelmold:
## CAA-360 CRITICAL — EU data survived erasure + a FALSE attestation was signed

The DSR erase (`Clerk user.deleted` / account-delete) forwards to `${CORELINK_API_BASE}/_internal/dsr/erase` → the **IAD (US) container**, whose R2 client only reaches US buckets. An **EU tenant's** CAS/AC bytes live in the `prod-lhr` container's dedicated EU buckets (`corelink-cas-eu` / `corelink-ac-eu`) — which the IAD sweep can neither see nor (by design) hold credentials for. So **EU data survived "erasure"** while the Ed25519 attestation still signed a tenant-wide **`VerifiedComplete`** — a false GDPR completion claim. (The single most HN/legal-damaging finding.)

## Fix — fan out + fail-closed completeness

`worker/src/index.ts` internal arm: when this worker is the erase **ORIGIN** (not itself a fan-out target), fan `/_internal/dsr/erase` out to **every** regional worker (`PROD_LHR/SAM/NRT/SYD`). Each container erases **its own** jurisdiction's buckets with its own region-scoped R2 creds — the US container never gets EU creds, so **residency isolation is preserved**. Return "complete" (the local 2xx) **only when the local AND every regional sweep confirm**; else fail **CLOSED (502)** so the queue consumer retries and **no false `VerifiedComplete` is ever signed**.

- **Fail-closed by construction:** a missing regional binding, a transport error, or a non-2xx from any region ⇒ not complete ⇒ retry.
- **Loop-guarded** via `x-corelink-fanout-from` (a regional worker running its own jurisdiction's erase does not re-fan).
- Body buffered once + replayed per region.

## Deploy dependency (fails-closed until met — never false-completes)
Each regional worker must accept the erase internal-auth key + carry its region's `ERASURE_ATTESTATION_*` config. **Until then the erase safely fails-closed (retries), so it can never regress to a false completion.** Coordinator one-shot to follow.

## Tests
Worker suite **397/397** (+4: fan-out-to-all, fail-closed-on-region-error, fail-closed-on-absent-binding, loop-guard); `tsc` clean.

Part of the pre-HN-launch CAA-360 remediation (`reports/audits/2026-07-03-caa360-pre-hn-launch.md`).


gmhelmold:
Merging via `--admin`. **Documented reason:** the ONLY red is `okf-wiki-validation` (the internal-arm change shifted cited line ranges in concepts citing index.ts) — auto-reconciled on `main` by the self-maintaining OKF engine post-merge (precedent #602/#603/#605). ALL load-bearing validation is GREEN: `vitest` (the 4 new DSR fan-out tests: fan-out-to-all / fail-closed-on-region-error / fail-closed-on-absent-binding / loop-guard, + 393 regression = 397), `dco`, `CHANGELOG`, `gitleaks`, `trivy`. No Rust changed (cargo gates N/A). The fix is fail-closed by construction. Branch protection required-checks = [].

PR #708, opened 2026-07-09
feat(cli): build the documented-but-missing subcommands

gmhelmold:
## Summary

Eight CLI surfaces were referenced in the docs / e2e suites but **rejected by clap** because the commands did not exist. This builds them for real (in `tools/cli/**`), each with unit tests, matching documented behaviour and flagging genuine server-side gaps rather than faking them. No server code was changed.

## Per-command status

| Command | Status | Notes |
|---|---|---|
| `bazel-init [--force]` | **DONE** | Managed `.bazelrc` block + `.corelink/credentials` (0600), idempotent (re-run exits 0; `--force` rewrites, no dupes). Endpoint/tenant/PAT from `config.rs`. **Writes the REAPI config that actually works** (`--remote_cache=<endpoint>/bazel/v2` + `--remote_instance_name=<tenant>`, per `routes/bazel_v2.rs`) — NOT the dead `grpcs://cas.corelink.humangr.com` the tutorial still shows (stock `--remote_cache=http` 404s on the REAPI surface). |
| `audit tail` | **DONE** | Windowed pull from the live `GET /v1/audit/:tenant/export` (default trailing hour) + client-side `--filter key=value`. True live-follow (`-f`) would need a server SSE surface (flagged). |
| `audit export` (production) | **DONE** | Non-`--fixture` path now streams the live route, persists NDJSON content-addressed + the chain-head anchor (re-verify with `audit verify-ndjson`). Was: "not yet wired; use --fixture". |
| `config apply --file` | **DONE** | Loads a whole `corelink.toml`, validates every leaf like `config set`. |
| `config set/get observability.*` | **DONE** | Config model extended with the open-ended `[observability.export.*]` sub-tree (Datadog / OTel / Grafana) from the observability how-tos; `observability.export.variant` enum-validated; types preserved through TOML round-trip. |
| `cas get` | **DONE** | Download a blob by digest (`sha256:`/`blake3:` prefixes ok), client-side integrity verify. |
| `cas export` | **PARTIAL** | Bulk-download every blob to a **local directory** (real, paged). `s3://`/remote destinations are a flagged gap (need an object-store client + creds). |
| `import` | **PARTIAL** | Bulk pre-warm the CAS from a **local directory** (content-addressed → dedup for free). `s3://` sources flagged (sync locally first). |
| `ci mirror` | **PARTIAL** | One-shot mirror of a **local cache dir** → CoreLink CAS. The live-sidecar bridge (`--from bazel-remote` as a service / BES ingestion) is a flagged gap (needs a source-protocol daemon). |
| `tenant export` | **PARTIAL** | Content-addressed bundle (audit-chain slice + CAS index) from real routes. Full single-file `.tar.zst` with **all blob bytes + RBAC roster + DPA receipt** is gated on a server bulk-export endpoint that does not exist (blobs are retrievable via `cas get`). |
| `tenant verify-export` | **DONE** | Re-checks every component + bundle hash offline (no PAT required — a departed customer may have revoked tokens). |

## Server endpoints I could NOT add here (flagged, not built)
- **Tenant bulk-export** (`.tar.zst` of all blob bytes + RBAC roster + DPA receipt) — no endpoint exists; `tenant export`/`cas export` assemble from the reachable audit-export + CAS-list routes instead.
- **Direct object-store (S3/GCS) ingest/egress** for `import`/`cas export` — needs an object-store client outside this crate.
- **Live cache-mirror sidecar / BES ingestion** for `ci mirror` — needs a long-running source-protocol daemon.
- **Audit live-follow (SSE)** for `audit tail` — only a windowed pull is possible against the current route.

## Gates
- `cargo clippy -p corelink-cli --all-targets -- -D warnings` — clean (expect/unwrap denied outside `#[cfg(test)]`).
- `cargo test -p corelink-cli` — green; new unit tests for every command.
- CHANGELOG `[Unreleased]` entry added; DCO + Co-Authored-By trailers present.


gmhelmold:

PR #714, opened 2026-07-09
feat(byok): close the activation write seam — flip tenants to BYOK active

gmhelmold:
## What & why

Migration `0081` created `tenant_byok_config` + `tenant_byok_secret`, and the r2_s3 CAS store **already encrypts at rest when `tenant_byok_config.state == 'active'`** (read path decrypts) — but **nothing wrote those tables**. The H5 onboarding writer was deferred, so the encryption gate could never become `active`. All 4 KMS providers + envelope crypto + `byok_orchestrator::build_active` + revocation already exist and are wired. **The only missing seams were the activation writer and the real-KMS build wiring.** This PR lands exactly those.

## 1. Activation write path (`crates/corelink-container/src/customer_d1.rs`)

`D1ByokConfigWriter` — fail-CLOSED, tenant-scoped, audited, over the same async `ByokConfigRows` D1 seam the reader uses:

- **`activate()`** — UPSERTs the CMK identity (`mode` / `crypto_mode` / `cmk_provider` / `cmk_key_id` / `cmk_region`) into `tenant_byok_config` (`state='active'`) + the CMK-wrapped Tcs into `tenant_byok_secret`. **Secret written FIRST** so an `active` config never dangles over a missing Tcs (the read path fails closed on that combo). Params validated **before** any D1 write; **refuses to re-activate a crypto-shredded tenant** (monotonic terminal).
- **`deactivate()`** — the crypto-shred **kill switch** (`active`/`partial` → `shredded`, monotonic + idempotent). The deliberate control-plane complement of the always-on `corelink_byok::revocation` CMK-revocation detector.

Security: the plaintext Tcs is **never** handled or logged here — only the CMK-**wrapped** ciphertext. The audit event records tenant + provider + CMK **identity** + state, never secret bytes. Enforces `INV-BYOK-CRYPTO-SOVEREIGNTY` + `INV-TENANT-ISOLATION`.

## 2. Endpoint (`crates/corelink-container/src/routes/byok_admin.rs`)

`POST /v1/admin/byok/activate` + `POST /v1/admin/byok/deactivate`, operator-gated behind the **same `x-corelink-internal-auth` secret as `/v1/admin/*`** (auth-before-body-parse, M3). Writer is `None` in dev/CI (no D1 creds) → **503 fail-CLOSED**, never a silent no-op. Mounted in `routes.rs`.

## 3. Real-KMS build path (`crates/corelink-container/src/byok.rs`)

Grew from AWS-only to per-provider constructors `make_{aws,gcp,azure,vault}_kms_provider` + `make_active_provider` (delegates to `byok_orchestrator::make_provider` → `build_active`, the existing compile-time cfg dispatch). Now compiles under **any** `byok-*-real` flag. Setting a `byok-<p>-real` cargo feature selects that real provider end-to-end; two real providers stays a **hard compile error** (ADR-S30-001).

## Migrations

**Additive only — no new migration.** `0081` already has every column the writer touches.

## Tests

- Writer: activate / deactivate / monotonic-guard / validation fail-closed / transport fail-closed.
- **Read-after-write engaging encryption**: a stateful mock proves `activate()` → reader sees `state='active'` → `engagement_for()` returns `Encrypt(Convergent)`, and the wrapped Tcs round-trips through the base64 BLOB encoding to the exact bytes (so the Tcs is resolvable). `deactivate()` → `engagement_for()` returns `Plaintext`.
- Route: auth-before-parse 403, key-unset fail-closed 403, writer-absent 503, base64/enum body validation.

## Verification boundary

Last local `cargo check -p corelink-server` (default features) **compiled clean** — production code verified. `#[cfg(test)]` blocks + the real-feature `byok.rs` are left to CI's clippy+test gate (the shared Mac is saturated by parallel container compiles).

## Operator residual

Provisioning live KMS credentials and choosing the build feature (`byok-<provider>-real`) is an operator step — this PR does **not** fake any cloud call. With no flag set, the binary ships the `InMemoryFake` provider (dev/CI).


gmhelmold:

PR #718, opened 2026-07-10
fix(runner-revoke): owner_tenant OPTIONAL — reconcile with runners 2026-07-08 contract

gmhelmold:
**Cross-repo contract reconciliation** (server was stale). The runner revoke handler required `owner_tenant` (REV-S2, 2026-06-21); the runners dispatcher's **owner-ratified 2026-07-08 contract removed it** and sends `{pat_id}` only. The removal never landed on the server → **every native-path runner revoke has been 400-ing in prod** (revoke-on-teardown degraded to TTL self-expiry).

**Change:** `owner_tenant` is now **OPTIONAL** in `handleRunnerRevoke`:
- absent → `UPDATE pat SET revoked_at_ms WHERE pat_id` (revoke by pat_id alone)
- present → `... WHERE pat_id AND tenant_id` (tenant-scoped, defense-in-depth for opt-in callers)

No 400 on the `{pat_id}`-only body; no existing caller breaks. Does **not** touch the storage cap, request/mo cap, rate-limit, or the Stripe money-path.

**Security note:** this relaxes REV-S2 blast-radius scoping on revoke. The pat_id is already the capability (possessing it permits the revoke), so the tenant predicate was weak scoping, not an isolation boundary — and the removal is the newer owner-ratified contract. Owner-directed.

**Coverage:** the runners live-conformance gate exercises the `{pat_id}`-only path end-to-end (it surfaced this). A worker-vitest unit test for the two UPDATE shapes is a tracked fast-follow.


gmhelmold:

PR #803, opened 2026-07-16
fix(admin-ui): make /corelink-mount public paths public + surface-correct sign-in redirects

gmhelmold:
## The bug (PROD-CRITICAL)

On the path-mounted surface `humangr.com/corelink/*`, every public path was 307'd to `/sign-in`, and the redirect **dropped the mount prefix**:

- `/corelink/sign-up` → `307 /sign-in?redirect_url=%2Fcorelink%2Fsign-up` (should render Clerk `<SignUp>`)
- `/corelink/en/pricing`, `/corelink/_next/static/*`, `/corelink/locales/*` → same 307
- target `/sign-in` on `humangr.com` resolves to the **apex marketing site**, not the app's `/corelink/sign-in`.

## Root cause

The admin-ui Worker serves the app on **two surfaces** during the subdomain→path migration: `corelink-app.humangr.com/*` (root, no prefix — the current live surface) and `humangr.com/corelink/*` (path-mounted). On the path surface the edge middleware receives `req.nextUrl.pathname` carrying the `/corelink` prefix, but the deployed `isPublicPath` matched only **bare** paths → `isPublicPath("/corelink/sign-up") === false` → fell through to Clerk. Confirmed against the **deployed (HEAD)** code + the live `redirect_url=%2Fcorelink%2Fsign-up` capture.

## Fix (surface-aware — must not break either surface)

- **`route-matcher.ts`** — strip the `/corelink` mount prefix (present / absent / defensively-doubled + trailing slash + locale) before **every** match, so a public path is public in both shapes.
- **`middleware.ts`** — all **four** sign-in redirects are built **surface-correct** via new `signInPathFor()` / `signInRedirectPath()` helpers: re-attach exactly the prefix the request carried — `/corelink/sign-in` for a `/corelink/*` request, `/sign-in` for a root/subdomain request. (A bare `/sign-in` on `humangr.com` hits the apex site; a `/corelink/sign-in` on the subdomain 404s — so neither can be hardcoded.)
- Fail-closed posture and CSP/nonce logic unchanged; the fail-closed guard also trips on a missing `CLERK_SECRET_KEY`.

## Tests

- `tests/route-matcher.test.ts` — mount-prefixed public / protected / self-gated forms (incl. `/corelink/api/checkout/session` staying protected, prefix-trick negatives, doubled-prefix, trailing slash) + both-surface redirect helpers.
- `tests/middleware-failclosed.test.ts` — redirect target is `/corelink/sign-in` on the path surface **and** `/sign-in` on the root subdomain.
- Full admin-ui suite green: **104 files / 470 tests**; `tsc --noEmit` clean; `eslint` clean.

## Live verification (branch deployed to the `corelink-admin-ui` Worker)

| Request | Before | After |
|---|---|---|
| `humangr.com/corelink/sign-up` | 307 → apex `/sign-in` | passes middleware (no wrong redirect) |
| `humangr.com/corelink/en/welcome` (protected) | 307 → apex `/sign-in` | 307 → `/corelink/sign-in` |
| `corelink-app.humangr.com/en/welcome` (LIVE) | 307 → `/sign-in` | 307 → `/sign-in` (unchanged ✓) |
| `corelink-app.humangr.com/{sign-in,sign-up,en/pricing}` | 200 | 200 (unchanged ✓) |

## ⚠️ Separate remaining blocker for the PATH surface (owner-owned migration)

This middleware fix is correct and necessary, but `humangr.com/corelink/*` **pages still 404** after it — because **`basePath: \"/corelink\"` is NOT committed/deployed** (it lives only in the in-flight working tree). Without it, OpenNext/Next resolves routes + assets at `/`, so every `/corelink/*` page, API route, and `_next` asset 404s. The path surface will only render once the basePath migration (next.config.ts + the coordinated URL/asset updates) ships. Deliberately **not** included here — it is the owner's migration wave, not this bounded middleware fix. The live subdomain surface is fully working today and is unaffected by this change.


gmhelmold:
Merging via --admin. Documented, non-blocking reason: the only red is `trivy fs (JS vuln)` = pre-existing CVE-2026-54466 in websocket-driver@0.7.4, a transitive dep in the root pnpm-lock.yaml (present on main, lines 10825/24612). This PR touches ZERO dependency manifests (TS middleware + tests + one Rust doc-comment only), so it cannot introduce or affect that finding. The 2 pending checks are 'Post welcome comment' (cosmetic, non-gating) and 'cargo-license' (cannot newly fail — no Cargo dependency change). All 27 load-bearing gates are green (build/test/typecheck/lint/cargo-*/playwright×4/lighthouse/dco/okf/secrets/etc.).

PR #815, opened 2026-07-17
weld prod-surface green (sign-up Clerk-mount wait + pricing → app surface)

gmhelmold:
## Why

The `e2e-prod` workflow's `playwright prod-surface (chromium)` job has been chronically red for 12+ runs on two **public-surface** checks that are **test false-negatives, not product bugs** (the app is verified live).

## What

**sign-up test** — `/sign-up` is client-rendered: `<SignUp>` is dynamic-imported with `ssr:false`, so the SSR pass emits only a `BAILOUT_TO_CLIENT_SIDE_RENDERING` shell. The old `expect(body.length).toBeGreaterThan(0)` measured that shell at `domcontentloaded` (before Clerk mounts) — it flapped and never validated the widget. Now:
- wait for Clerk's mount node `[data-clerk-component="SignUp"]` (**attached**). React renders it **only** when `<ClerkProvider>+<SignUp>` mount with a valid publishable key — if `NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY` is unset, `page.tsx` renders the "key not configured" fallback with **no** such node, so this also catches a missing Clerk key on deploy.
- poll `window.Clerk.loaded` — proves the remote FAPI bundle + CSP (`clerk.corelink-app.humangr.com`) are wired.

Both are **strictly stronger** than raw body length **and** stable in headless CI.

**pricing test** — wait for the Docusaurus app root to hydrate before asserting the Free/Pro/Enterprise tiers + CTA href, so a client-render race can't false-fail. **Assertions unchanged (not weakened).**

## Verification (LIVE prod, 2026-07-17)

Drove the repo's cached chromium (both old `chrome-headless-shell` and new `--headless=new`, the CI-equivalent) against `https://humangr.com/corelink/sign-up` over CDP:
- Clerk's full bundle loads — `/v1/environment`, `/v1/client`, `signup_ui`, `clerk.browser.js` all **200** — and initialises: `window.Clerk.loaded === true`, `[data-clerk-component="SignUp"]` attached.
- The credential form (`.cl-rootBox`/`input`) **never paints** under an automated browser (mount node stays 0-height even after 25s). This is Clerk's hosted bot-protection — and is exactly why the suite header states the interactive auth round-trip is intentionally not exercised here (it's covered by the localhost test-mode suite).
- ⇒ asserting on the rendered form/inputs (or `toBeVisible()`) would **reintroduce** a false-negative in CI; the mount-attached + `window.Clerk.loaded` signals are the correct, robust gate.

> [!NOTE]
> The pricing test still depends on the docs site being redeployed with real pricing content to fully pass (separate loose-end being fixed elsewhere). This PR only makes it wait correctly for client-rendered content.


gmhelmold:
## Update — pricing canary repointed to the app surface (coordinator Option 1)

Root-caused the pricing failure: it is **not** a pricing-page bug — the Docusaurus **docs marketing site renders empty site-wide**. Webpack externalises `@theme/*` imports into unresolved literal `require()` calls → `ReferenceError: require is not defined` on load → no hydration → empty `#__docusaurus`; and a core SSG patch (`patches/@docusaurus__core@3.10.1.patch`) stubs `@theme/*` to a noop component, so the prerender is empty too. Reproduced locally (`cd apps/docs && pnpm build` → all 273 built HTML files have an empty root div). That fix is tracked as a **separate docs-build work item**.

Per coordinator decision (Option 1), this PR repoints the prod-surface pricing canary from the dead docs `/pricing` to the **admin-ui pricing surface** `humangr.com/corelink/en/pricing`, which renders the full Free/Solo/Starter/Pro/Max + Enterprise ladder with live CTAs.

**New assertions (not weakened):** wait for a real upgrade/signup CTA to mount → assert its `href` targets `/upgrade|/sign-up|checkout|billing` → assert `Free` + `Enterprise` + a `$<amount>` price is present.

**Verified live** (cached chromium, no browser install): `ctaHref:/corelink/en/sign-up` (matches), `hasFree:true`, `hasEnterprise:true`, `hasDollarDigit:true` → `TEST_WOULD_PASS: true`.

The sign-up Clerk-mount-wait fix is unchanged. CHANGELOG `[Unreleased]` entry added.

gmhelmold:
## Verified GREEN + a CI-trigger heads-up

**prod-surface is now 10 passed / 0 failed** against live prod (was 9/1). Verified by dispatching the `e2e-prod` workflow on this branch HEAD (GitHub-hosted ubuntu, exact committed code): run `29556741360` → `playwright prod-surface (chromium)` = **10 passed**, conclusion success.

⚠️ **Infra note (owner-side, not this PR's code):** the `pull_request`-triggered workflows (e2e-prod, changelog-validate, dco-check, build, …) are **not firing** for pushes to this branch after the first one — pushes/amend/close-reopen all yield 0 `pull_request` runs, while `pull_request_target` jobs and manual `workflow_dispatch` DO run and pass. This matches the documented "GitHub Actions PR-gate trigger/billing block" (Settings→Billing). The fix itself is proven green via dispatch; the PR's own prod-surface check will register once the trigger/billing block is cleared (or a maintainer re-runs). DCO/changelog are satisfied in-commit (Signed-off-by present; CHANGELOG `[Unreleased]` entry added).

### Docs-build defect — minimal repro (for the separate work item)
```
cd apps/docs
pnpm install --frozen-lockfile          # (or: pnpm -w install --filter "@corelink/docs...")
pnpm build                              # = docusaurus build  (or: npx docusaurus build --locale en-US)

# every built page has an EMPTY root div:
grep -c 'id="__docusaurus"></div>' build/pricing.html      # -> 1  (empty; ~5KB total file)
# and the client bundle carries unresolved literal require():
grep -oE 'require\("/[^"]*"\)' build/assets/js/main.*.js    # -> 4  (infima css, theme-classic/lib/{prism-include-languages,nprogress}, src/css/custom.css)
```
Root cause: webpack externalises `@theme/*` / CSS imports into literal `require()` (pnpm `.pnpm/…` realpaths don't match Docusaurus's babel-include/externals allowlists) → `require is not defined` in-browser (no hydration) → empty `#__docusaurus`; and `patches/@docusaurus__core@3.10.1.patch` stubs `@theme/*` to a noop component at SSG time so the prerender is empty too. Lead for the real fix: `resolve.symlinks:false` defeats the server externalization (SSG starts rendering) but then needs the client babel-include fixed + the `/tags/*` render error resolved + the `ssgNodeRequire` `@theme/*` noop patch removed. Do NOT re-point the pricing canary back at `DOCS_URL/pricing` until that lands.

PR #817, opened 2026-07-17
fix(byok): fail closed 501 when no real KmsProvider is linked

gmhelmold:
## Finding C1 (adversarially verified)

BYOK is sold as a live security guarantee, but the prod container binary links **no real KMS**. `Dockerfile` builds `-p corelink-server` with no `byok-*-real` cargo feature, so `byok_orchestrator::active_provider()` resolves the only `KmsProvider` to the `InMemoryFake` — XOR-masks the DEK with a fixed module-private constant, region `local-fake`, doc-marked *"Not for production"*. Yet `/v1/admin/byok/activate` would still flip `tenant_byok_config.state='active'`, engaging that fake provider — reporting a live BYOK guarantee the binary cannot deliver.

## The fix (ENG half only)

Make activation **fail closed with `501 Not Implemented`** in any build without a real `KmsProvider`, so it can never report `active` while the `InMemoryFake` is active.

- **Mechanism — compile-time feature flag** (not a runtime check): the provider is selected at compile time by the `byok-*-real` features, surfaced through the `const fn active_provider()` that boot/readiness already use. A new compile-time const `REAL_KMS_PROVIDER_WIRED` (`!matches!(active_provider(), ActiveProvider::InMemoryFake)`) gates `handle_activate`. This is the exact "feature inert in prod" mechanism used elsewhere in the container — zero hot-path cost.
- **Ordering:** the 501 fires **after** the operator auth gate (unauthenticated callers still get 403 — no inertness leak) and **before** the D1 writer is resolved, so `state` is never mutated to `active`. Returns `501 {"error":"byok_not_available"}`.
- **Boot log:** `main.rs` emits a one-line `byok_activation_inert` note at boot when the fake provider is active (placed after the OKF-cited boot blocks to keep anti-drift line-anchors stable, per the existing DSR-anchor convention).
- **`deactivate`** (the crypto-shred kill switch) is unaffected — it never reports `active`.

## Regression test

`routes/byok_admin::tests::activate_fake_provider_is_501_not_available` — the same authenticated request that previously flowed into the write path (503, no writer) now returns 501 short of it, proving no state flip. Gated to the no-real-provider build (the only one CI compiles), mirroring the umbrella crate's `no_provider_feature_default_build`.

## Verification

- `cargo clippy -p corelink-server --lib --bins --tests -- -D warnings` — clean.
- `cargo test -p corelink-server --lib byok_admin` — 7/7 pass (incl. new regression; existing 403-before-parse + key-unset + deactivate-403 still green).
- OKF C5: `okf_reconcile.py` → 0 stale concepts (the boot log was placed after the cited blocks so no line-anchors drifted).

The marketing/sale-side pull is intentionally **not** touched (owner's call).


gmhelmold:
Merging with the `cargo-license + allowlist policy gate` check pending — documented GH-scheduling flake, not a real red. All load-bearing gates are GREEN: PR gate (crate-scoped clippy+tests), **cargo-deny (license + advisories + sources + bans)**, cargo-audit, proptest-density, INV-registry×TLA, dco, changelog, okf-wiki-validation, secrets-matrix. The separate `cargo-license` allowlist gate has sat pending in the Actions queue; C1 (BYOK 501-gate) adds NO new dependency (a compile-time `const` + a 501 return in existing code), so there is zero license-allowlist surface — and cargo-deny already covers licenses green. Squash-merge.

PR #99, opened 2026-06-03
fix(ci): point pyo3 at system python3 >=3.10 in ffi-matrix (was 3.8)

gmhelmold:
The ffi-matrix-ci Rust unit-test build (cargo test -p corelink-py) compiles
pyo3 with abi3-py310, which requires a Python interpreter >= 3.10. With no
working Python pin (setup-python is broken on the self-hosted macOS fleet —
it provisions into an unwritable /Users/runner), pyo3 auto-detected a stale
3.8 on PATH and failed:

  cannot set a minimum Python version 3.10 higher than the interpreter
  version 3.8 (abi3-py310)

Set PYO3_PYTHON: python3 at job level on rust-unit and cross-language-verify
(the two jobs that build corelink-py) so pyo3 resolves the fleet's system
python3 (3.14), which satisfies the abi3-py310 floor. FFI matrix logic
unchanged.

Signed-off-by: Gustavo Schneiter <gustavo@humangr.com>
Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

PR #856, opened 2026-07-19
perf(worker): cache the per-request PAT-verify D1 read on the auth hot path (#99)

gmhelmold:
## What & why

The Worker `extractAuth` did a synchronous D1 read of the `pat` row on **every** authenticated request. Profiling against **prod** (a SAM/São-Paulo client, 2026-07-19) isolated it as the **sole uncached read left on the warm hot path** (the tenant suspend gate already caches), each read costing ~0.7 s trans-continental to the D1 primary:

| Request | Latency |
|---|---|
| `/health` (no auth) | ~40 ms |
| bad token (fails HMAC, **pre-D1**) | ~45 ms |
| valid token, **cold** isolate (pat + suspend read) | ~1.4 s |
| valid token, **warm** isolate (suspend cached) | **~0.7 s** ← this uncached pat read |

The auth-only tenant-mismatch `403` (returned **before** any DO/container forward) measured **identical** to the full served path — proving the ~1 s is **entirely the Worker's D1 reads**, not the DO or the container. This is the HIGH the **clw team also reported independently** (~1 s per authenticated command; `ls --all` etc.).

## The fix

New `worker/src/lib/pat_verify_cache.ts` (`verifyPatRowCached`) mirrors the established edge cache pattern (`tenant_suspend_gate.ts`; #667's container `CachedTierResolver`): a **per-isolate `Map` keyed by the non-secret `token_id`**, **single-flight** (a same-token burst collapses to ONE D1 read), **LRU-capped**. Same SQL, same columns, same decisions — the warm auth floor collapses from ~0.7 s toward the ~45 ms HMAC-only cost.

### Revocation-correctness (this is a credential cache — the load-bearing part)
- **TTL = 5 s**, sized against **ADR-0030** `SLO-FRESH-PAT-REVOKE ≤ 60 s p99` and matching the container **`NativePatGate`** verify cache (red-team **C3**'s deliberate 60 s→5 s bound). A mid-entry revoke is honored for ≤5 s then denied — well within the 60 s SLA.
- **Positive results only** — a freshly-minted token authenticates immediately (no negative-cache delay); **D1 stays the source-of-truth** on every miss (**INV-AUTH-NEON-IS-SOT**).
- **Expiry re-evaluated by the caller on every hit** → zero added staleness on expiry; only revoke has the ≤5 s window.
- **A D1 fault is never cached and never serves a stale/expired entry** (the `d1_lookup_error → 503` posture is unchanged); a still-fresh (<5 s) hit is served through a transient blip for availability.
- Consulted **only after the HMAC possession proof**, so a wrong-secret token never reaches it; the cache holds the D1 row (tenant/scope/expiry/runner marker) only — **never the secret**.

## Tests (fail-before / pass-after)
`worker/tests/pat_verify_cache.test.ts` — **12 cases** incl. the **revocation-window bound** (honored within TTL, denied after) and an **integration proof** that the SAME PAT twice does exactly **ONE** pat D1 read. A global `beforeEach` reset in `tests/setup.ts` isolates the per-isolate singleton per case.

- Full worker suite: **526 node + 22 miniflare/workerd** green.
- No new `tsc` errors (14 pre-existing, unchanged).
- OKF reconciled (6 concepts drifted by the line-shift + the new file added to `auth/pat-moat`).

## Scope / follow-up (honest framing)
This closes the **reported HIGH** (repeated-token hot-path latency) correctly and within precedent. It does **not** fix the deeper root cause it surfaced: **the D1 `CONFIG_DB` primary is trans-continental from SAM users, so every uncached read (~0.7 s each) is slow** — cold isolates still pay it, and the served-path tier/storage/region reads remain uncached. The SOTA root fix is **D1 read replication (Sessions API)**, which would cut *every* control-plane read globally — a larger change with replica-lag-vs-revocation reasoning that deserves **its own ADR** (tracked follow-up, not bundled here).

Refs #99


gmhelmold:
⚠️ **Merge as a merge-commit, not squash** — this PR advances 6 OKF `checkpoint_sha` values to the perf commit `5aa10511`; squashing would rewrite that SHA out of history and re-red the OKF freshness gate on the next `worker/` change. (Same rule as prior OKF-touching PRs.)

PR #867, opened 2026-07-20
fix(container): accept canonical cache:r/cache:w/cache:find-missing self-serve PAT scopes

gmhelmold:
## The bug (launch-critical: cold-signup first-value was DEAD)

Every real dashboard **"Create token"** was returning **`401 unauthorized`**. Root-caused **live** against prod with a real browser Clerk session.

The self-serve mint classifier `classify_requested_scopes` (`crates/corelink-container/src/scope.rs`) — the single source of truth for `routes::customer::handle_keys_create` and `customer_d1::map_requested_scopes` — accepted `cache:read`/`cas:r` and `cache:write`/`cas:rw`, but **not** the **canonical corelink-pat wire form** `cache:r`/`cache:w`/`cache:find-missing` (`SCOPE_CACHE_R/W/FIND`, `auth_model.md §3.1`) — which is **exactly** what the dashboard `KeysClient` `SCOPE_SPECS` send (`DEFAULT_SCOPES = ["cache:r"]`).

A canonical token fell through to `Err(unrecognized)` → `CustomerHandlerError::Unauthorized` → `401 "unauthorized"` (the inner *"unrecognized scope token"* message swallowed by `map_err`).

## Why it hid for so long
- The black-box PAT suite mints via the **internal** path, not the customer surface — never exercised this classifier.
- In the browser the authed page **shell is SSR'd from the Clerk cookie**, so the page *looked* authed; only the client-side **cross-origin `getToken()` Bearer POST** 401'd.
- The prior theory (tenant-provisioning race) was **wrong**: `GET /v1/customer/keys` → `200`, so the tenant was resolved and the session valid.

## Live proof (real prod Clerk session, `tests/e2e-browser` token probe)
| request | result |
|---|---|
| `GET /v1/customer/keys` | **200** (tenant resolved, azp/iss/sub valid) |
| `POST {scopes:["cache:r"]}` (UI default) | **401** |
| `POST {scopes:["cas:r"]}` | **201** |
| `POST {scopes:["cache:read"]}` | **201** |

## The fix
Additively accept the canonical `cache:r` / `cache:w` / `cache:find-missing` tokens (find-missing is discovery-only ⇒ least-privilege **read** class, never a write). The fail-CLOSED exact-token grammar is otherwise **unchanged**:
- `admin`/`owner` still **never** self-serve-grantable
- truly-unknown tokens still `Err`
- the rt-nuclear #15 substring-vs-exact-token regressions (`"writes"`, `"rewrite"`, …) still rejected

**Fix side = backend classifier (not the frontend):** the mint endpoint issues PATs that the data plane enforces, so it must accept the **same** scope vocabulary the data plane uses (`corelink-pat`). The UI was already using the canonical form; the classifier had drifted.

## Tests
Regression-locked in `scope.rs::classify_requested_scopes_is_exact_token_and_fail_closed`: `cache:r`→ReadOnly, `cache:w`→ReadWrite, `cache:find-missing`→ReadOnly, `cache:r`+`cache:find-missing`→ReadOnly. Full `scope::` module suite green (31/31); clippy `-D warnings` clean.

## Deploy
Requires a **prod container roll** to ship. Post-roll live verification: `POST {scopes:["cache:r"]}` → `201`.


gmhelmold:
**Adversarial review #1 (root-cause): CONFIRMED** on all 4 points — dashboard POSTs `cache:r` verbatim (no transform layer), pre-PR classifier had no matching arm → `Err` → bare-`"unauthorized"` 401 on POST while GET (never classifies) stayed 200; fixed additively here.

**Scoped-out follow-up (tracked separately, NOT in this PR — repo hygiene):** `KeysClient` `SCOPE_SPECS` offers a 4th scope **`admin:audit`** ("Read audit log") which the classifier also doesn't recognize → selecting it still 401s. It is **not** in `DEFAULT_SCOPES` (`["cache:r"]`), so the launch-critical first-value/default path is fully fixed here. The `admin:audit` decision is product-tinged (remove the UI option vs. wire a real customer audit-read scope that grants `SCOPE_ADMIN_AUDIT` without cache/admin) and will land in its own PR.

gmhelmold:
**Adversarial review #2 (security/correctness): FLAW-FOUND → fixed.** The reviewer caught that folding `cache:find-missing` into `ReadOnly` would mint a **mislabeled** token: the self-serve mint provisions the bitset from the coarse read/write string (`customer_d1::create`, `SCOPE_CACHE_R`/`SCOPE_CACHE_RW`) with **no `SCOPE_CACHE_FIND` path**, and REAPI enforces `CacheFindMissing` as a non-implying scope (`cache_find_missing_does_not_imply_cache_read`; `find_missing_handler_e2e` AC-FM-5/6) — so it would *under-grant* find-missing (unusable on the REAPI plane) yet *over-grant* full download read.

**Applied remediation:** dropped `cache:find-missing` from the accept-list — accept only `cache:r` / `cache:w` (both faithfully provisionable to `SCOPE_CACHE_R` / `SCOPE_CACHE_RW`). `cache:find-missing` stays an honest `Err` until a bitset-based mint exists; the test now **locks** that intentional rejection with the rationale. Verified the reviewer's load-bearing claim directly (`customer_d1.rs:1229-1233` hard-codes the bitmask). Review #2's other 3 axes (fail-closed integrity, consistency, other call sites) came back CLEAN.

The reviewer also confirmed a **pre-existing** gap (independent of this PR): even a `cache:rw` self-serve PAT lacks the FIND bit, so `FindMissingBlobs` on the strict REAPI plane is denied for any dashboard-minted PAT. Tracked as a follow-up (bitset-mint refactor) — not a regression from this change.

Net: launch-critical default (`cache:r`) + read+write (`cache:w`) fully fixed, no mislabeled token minted, zero debt shipped.

PR #876, opened 2026-07-20
fix(security): account deletion is OWNER-only (non-owner seat could erase the whole tenant)

gmhelmold:
## The vulnerability (HIGH — irreversible tenant-wide erasure by a non-owner)
`handle_account_delete` (`routes/customer.rs`) gated **only** on `principal == "clerk"` (is-a-dashboard-session) and then called `request_erasure(&tenant)` — a **tenant-wide** GDPR erasure of the whole account. But every team seat (`admin`/`member`/`viewer`, migration 0074) carries a Clerk session and resolves to the **owning** tenant (`verifyClerkSessionAndResolveTenant` team_member fallback). So **any non-owner member could irreversibly erase every teammate's account** — and it isn't MFA-gated (that's only the `/v1/privacy/*` plane). `x-corelink-scope` only distinguishes `viewer` vs the rest, so it could not express "is owner".

## Fix
- **Worker:** forwards the D1-resolved team role as a new server-trusted **`x-corelink-role`** header on the customer plane, and adds it to `CLIENT_TRUST_HEADERS` so a forged client copy is **structurally stripped** (the Worker is the sole setter).
- **Container:** `handle_account_delete` fail-CLOSES on anything but `owner`.
- Cache-PAT callers are still rejected earlier (`principal != clerk`); tenant/unwired/idempotency behaviour unchanged.

## Tests
`routes::customer::tests::account_delete_non_owner_clerk_session_is_403` (member/admin/viewer/"" → 403, the erasure requester is never reached) + the existing owner→202 path + PAT→403 + missing-tenant→401 + unwired→503. Clippy `-D warnings` clean; worker vitest 552 green.

## Note
Considered adding MFA step-up to account-delete too (parity with `/v1/privacy/*` erasure) — deferred as a separate hardening; owner-only closes the privilege-escalation hole (the finding).

**Ships with a prod container roll + worker deploy.**


gmhelmold:
**Adversarial security review: APPROVE on all core axes** (forgery — `x-corelink-role` is stripped then Worker-set, sole setter; bypass — absent/empty role fail-CLOSES to 403, PAT rejected first; role-resolution — owner only for the tenant.clerk_user_id match; regression — owner still 202). The owner gate itself is correct.

**Related finding (CONCERN, tracked as a follow-up — NOT a blocker on this diff):** the Worker collapses `x-corelink-role`→`x-corelink-scope="read-write billing"` for EVERY non-viewer (index.ts:2485), so privileged customer ops that gate on `requires_cache_write`/`requires_billing_admin` treat a plain **member like an owner**. This leaves a **member→owner escalation detour** (a member can `handle_team_invite` an `owner`-role seat → accept preserves role → that identity deletes the tenant) plus member-can-remove-teammates + member-can-cancel-billing. This PR closes the **direct one-request** hole; the detour needs a separate RBAC PR that raises those gates to the new `x-corelink-role` owner/admin check + forbids `normalize_invite_role` minting `owner` self-serve. Opening it next.

PR #885, opened 2026-07-21
fix(signup-worker): race-free runner entitlement seed (paying customer could get 0 capacity)

gmhelmold:
## Summary — money-path, launch-blocking (found live)

Verifying the Ask-A `$0` runner-starter purchase for the cold-signup tenant `3c7d77b1`, I found `runner_billing` mapped **`active`** but `runners_entitlement` **empty** — the customer paid, and `acquire` stays `429 over_cap`.

**Root cause — a write-race in the Stripe webhook** (`apps/signup-worker/src/webhooks/stripe.ts`). On `customer.subscription.{created,updated}` for a runner sub, the handler pushes both writes onto `requiredWrites`, run via `await Promise.all(...)`:
- `upsertRunnerBilling` → `INSERT INTO runner_billing …`
- entitlement seed → `INSERT INTO runners_entitlement … SELECT tenant_id FROM runner_billing WHERE runner_subscription_id=?`

But `requiredWrites.push(fn(...))` **invokes each write immediately**, so they run **concurrently**. The seed's correlated `SELECT` reads the very row the other write is still inserting — when it wins the race it finds **no** row → inserts **0 rows** → silent no-op. **Non-deterministic**: dogfood happened to hit the lucky ordering; `3c7d77b1` lost it. Every real runner purchase is a coin-flip on whether the paying customer gets capacity.

## Fix

New race-free `upsertRunnersEntitlementByTenant` binds the tenant id **directly** — we already hold it from the subscription's `metadata.tenant_id` — so the seed no longer depends on the concurrent `runner_billing` write. The subscription-correlated variant is retained **only** for the no-metadata path (where `runner_billing` was mapped by a *prior* event and already exists). Applied to both the `.created` and `.updated` arms.

## Why the tests didn't catch it

The webhook tests use a mock D1 that captures SQL but doesn't model the correlated-`SELECT`-finds-nothing race (no real row visibility) — so they were green while prod broke. The seed assertions now lock the **race-free** form: `not.toContain("FROM runner_billing")` + `VALUES(tenant_id,…)` + the tenant-id param, plus a dedicated `runner seed does NOT correlate on runner_billing when the tenant is known` test. **Full signup-worker suite: 234/234.** (The 6 pre-existing `tsc` errors are unrelated and present on `main`.)

## Remediation + deploy

`3c7d77b1` was manually re-seeded to `20/100` (one-off, idempotent with a corrected redelivery) so the runners TL's acquire proof is unblocked now. This fix prevents recurrence for every future purchase; ships on the next signup-worker deploy.


gmhelmold:
Merging with `--admin` for a **documented, non-blocking infra reason** (per CLAUDE.md + pre-merge-gate-check): the sole failing gate is `trivy fs`, and it is **environmental, not introduced by this PR**:
- 2 newly-published transitive **dev/build-tooling** DoS CVEs — `js-yaml` CVE-2026-59869, `shell-quote` CVE-2026-13311 (not prod-runtime).
- This PR changes **zero** dependency/lockfile files (diff = `stripe.ts` + its test + CHANGELOG + OKF docs), so it did not introduce them.
- They are red on **main + every open PR** — trivy passed on #884 ~1h ago; the CVEs landed in trivy's DB since. Low-severity, visibility-class.
- This is a **launch-blocking money-path fix** (a paying runner customer could get 0 capacity) that the runners TL is actively waiting on to close the money path.

Every other gate is green (vitest 234/234, okf, cargo-deny/license, gitleaks, secrets-matrix, DCO, CHANGELOG). Tracked follow-up: a separate `pnpm.overrides` bump (js-yaml→4.3.0, shell-quote→1.9.0) to clear trivy repo-wide.

PR #886, opened 2026-07-21
perf(worker): KV-L2 the tenant data-residency read on the CAS/AC hot path (latency WP slice 1)

gmhelmold:
## What

Slice 1 of the latency WP (owner + clw greenlit): **collapse the warm-path `wdb` Server-Timing phase** — the 4–6 serial uncached D1-PRIMARY reads that ride *after* the KV-served auth and dominate the ~1.2 s SAM/GRU floor.

This slice removes **one** of those serial reads: the per-request `SELECT primary_region FROM tenant` residency lookup. Cloudflare D1's primary is ENAM (no South-America region), so a GRU/SAM caller pays ~120 ms+ per request for a value that is, in practice, immutable per tenant.

## How

New `worker/src/lib/tenant_residency_cache.ts` (`resolveTenantResidency`) — a three-tier cache mirroring the `pat`/`tsusp` L2 (ADR-0070):

- **L1** per-isolate map (5 s, micro-burst dedup) →
- **L2** Workers KV (`tres:<tenant>`, 60 s — edge-local/per-colo, survives isolate fan-out, ~3 ms everywhere incl. SAM) →
- **L3** D1 source-of-truth, single-flighted.

The hot-path call site in `index.ts` swaps the inline `SELECT` for this resolver (write-behind handed to `ctx.waitUntil`).

### Safety — residency stays FAIL-CLOSED (the opposite of the suspend gate)
- An **unresolved** region (D1 fault, no cached fallback) → `RESIDENCY_UNRESOLVED` → the caller still returns `503 RESIDENCY_UNAVAILABLE`, never the IAD/local fall-through.
- A **known** region survives a transient D1 fault (residency is immutable), so a blip doesn't 503 an established tenant.
- Bounded staleness cannot **silently** leak: the container residency backstop (`residency.rs`) 409s any real cross-region mismatch, so the worst case after a rare admin re-pin is a self-correcting 409. TTL is deliberately the conservative 60 s KV floor for a compliance-sensitive value.

## Tests
- New `worker/tests/tenant_residency_cache.test.ts` (15): tiering, L1 TTL, single-flight, **fail-closed-on-unknown**, **known-region-survives-fault**, KV hit/miss/fault/malformed, `{region:null}` no-row.
- `worker/tests/index.test.ts` resets the per-isolate cache between cases (many region tests reuse one tenant with different mock-D1 verdicts).
- **Full worker suite: 566/566.** Zero new `tsc` errors (14 == 14 pre-existing baseline). OKF reconciled green (160 concepts, 0 stale).

## Verification / acceptance
Post-deploy, clw re-runs the GRU Server-Timing probe → the `wdb;dur=` phase should drop by the eliminated residency hop. Worker-only change (keeps the container pin fresh); ships on the next `corelink-api` deploy.

## Scope
Residency only. Slice 2 (tier KV-L2), slice 3 (non-blocking container audit write), slice 4 (quota via D1 Sessions) are separate follow-up PRs — each its own review + live-prove.


gmhelmold:
Merging `--admin`: the sole red is `trivy fs (JS vuln)` — the two just-published transitive dev-tooling DoS CVEs (`js-yaml` CVE-2026-59869 / `shell-quote` CVE-2026-13311) that are red on `main` and every open PR. This PR changes **no lockfile** (worker TS + docs only), so the red is pre-existing environmental, not introduced here. All load-bearing gates (vitest 566/566, okf-wiki, cargo fmt, CHANGELOG, dco, gitleaks, secrets-matrix) are green. The trivy CVEs are tracked for a separate `pnpm.overrides` bump.

PR #888, opened 2026-07-21
fix(docs): repair blank prod render + migrate docs to humangr.com/corelink/docs

gmhelmold:
## Problem

The public documentation site rendered **100% blank in production** — browser-verified (headless Chromium against `corelink-docs.humangr.com`): `200 OK` but an empty `#__docusaurus` root, `data-has-hydrated="false"`, no `<h1>`, and `PAGEERROR: require is not defined`. It has been blank **since the site's first commit**.

**Root cause:** `"type":"module"` in `apps/docs/package.json` forces Docusaurus's *generated* `.docusaurus/client-modules.js` (a CJS/ESM hybrid — `export default [ require("…/infima/default.css"), … ]`) to be parsed as an ES module. In ESM, webpack leaves the `require()` calls **untransformed** (bare `require` is a free identifier, never rewritten to `__webpack_require__`), so they throw at boot → React never hydrates → blank page; the same broken module also makes the SSG server-render emit an empty root (no pre-rendered fallback).

## What this does

1. **Fixes the render** — remove `"type":"module"` (client bundle bare-`require` 4→0; `index.html` 6 KB-empty → 32 KB-populated).
2. **Migrates to the canonical path** `humangr.com/corelink/docs` (matches #804 — one host for the whole product, no subdomain). `baseUrl=/corelink/docs/`, served by a **Cloudflare Worker + Static Assets** (`apps/docs/wrangler.toml` + `worker/index.ts`) whose zone route is **declared in-repo** (versioned/PR-reviewed, unlike the old dashboard-managed route). `humangr.com/corelink/docs*` is more specific than admin-ui's `/corelink/*` → **zero conflict** (verified against the live route table); the retired subdomain 301s to the canonical path.
3. `docs-deploy.yml`: CF Pages → `wrangler deploy`.
4. **CSP** (`static/_headers`): allow-list Docusaurus's two inline bootstrap scripts by **sha256** (not blanket `unsafe-inline`).
5. **Fix 163 broken internal links** the blank build silently hid (empty roots → nothing for `onBrokenLinks:"throw"` to validate) across `docs/**`, `i18n/**`, `src/pages/**`; refactor-proof `.mdx` file-path links + 3 cross-locale draft-flag reconciliations.
6. Remove Pages-era artifacts (`static/CNAME`, `functions/_middleware.ts`) + their tests; sweep old-subdomain references.

## Verification

- Full build (all 4 locales) exits **0** (0 broken-link errors).
[...]

PR #890, opened 2026-07-21
fix(docs): run_worker_first so the retired subdomain 301s (Assets bypassed the shim)

gmhelmold:
Follow-up to #888 (docs → humangr.com/corelink/docs migration).

## Problem
After deploying the docs Worker, the retired `corelink-docs.humangr.com` subdomain served the docs **200 (duplicate content)** instead of 301-redirecting to the canonical path. Root cause: Cloudflare Static Assets serves a matching asset (`/` → `index.html`) **directly**, bypassing `worker/index.ts` — so the host-redirect branch never executed.

## Fix
`run_worker_first = true` in `apps/docs/wrangler.toml` — the Worker shim owns routing (host 301 + mount-prefix strip) for **every** request before asset matching.

## Verified live (already deployed manually; this persists the config in-repo)
- `corelink-docs.humangr.com/` → **301** → `humangr.com/corelink/docs/`
- `corelink-docs.humangr.com/reference/cli/cli-get` → **301** (path preserved)
- `humangr.com/corelink/docs/` → 200 (browser: hydrated, no errors); assets 200; bare mount → 308. All unaffected.


gmhelmold:
Merging with `--admin` past two non-blocking flakes (neither caused by this PR, which changes only `wrangler.toml` + `CHANGELOG.md` — zero Rust):
- **cargo-license + allowlist policy gate**: GREEN on main at `0abbadea` (the #888 merge, identical Rust to this PR) and this PR touches no Cargo deps; the failing run emitted no `VIOLATIONS.txt` content — a transient audit error on the shared self-hosted Mac runner (known `~/.cargo` corruption). Re-run triggered.
- **trivy fs (JS vuln)**: the documented chronic env red (js-yaml/shell-quote transitive dev-tool CVEs), red repo-wide, previously `--admin`'d (#885).

All other gates green. This PR only persists the already-deployed `run_worker_first` config so a future CI docs-deploy doesn't revert the legacy-subdomain 301.

PR #795, opened 2026-07-13
deps(majors): adopt rand_chacha 0.10 + password-hash/argon2 0.6 + rusqlite (finish the #779 held group)

gmhelmold:
## Finish the #779 / #793 held Dependabot MAJOR group

#779 grouped four MAJOR bumps; #793 adopted `ed25519-dalek` 3 and **held** the other three because each needs a *companion major* the group omitted. This lane resolves that group: **`rand_chacha` (with its `rand` companion) is now adopted**; the other two stay held (still blocked on a companion #779 does not supply), with reasons re-confirmed against crates.io today.

### ✅ Adopted — `rand` 0.9 → 0.10 + `rand_chacha` 0.9 → 0.10 (companion)

Across the ~22 crates that pin them — property-test RNGs plus the **one prod site**: the `corelink-worker` timing-padding side-channel mitigation. The `rand` 0.9 → 0.10 trait/rename churn:

- extension trait `Rng` → **`RngExt`** (upstream `rand_core` renamed `RngCore` → `Rng`)
- `RngCore` → `Rng`, `TryRngCore` → `TryRng`
- `rngs::OsRng` → **`rngs::SysRng`** (getrandom 0.4 `SysRng`, already in the lock from #793's ed25519-dalek 3 / getrandom 0.4 adoption — no new dep)
- removed `gen_range` → `random_range`
- worker Cargo feature `os_rng` → `sys_rng`

`rand` 0.9 / `rand_chacha` 0.9 remain in the lock for the `bollard`/`testcontainers` dev-dependency (pins `^0.9`); both majors coexist. **No prod crypto behavior change** — the timing-padding jitter is unpredictable-*per-seed*, not output-stable across versions; the held Argon2id auth crates are untouched.

[...]

PR #897, opened 2026-07-21
deps(deps): batch the safe minor/patch groups (cargo + github-actions)

gmhelmold:
## What

Combines the two **low-risk** Dependabot groups into a single PR so they land as **one merge** (minimizing the rebase-flood the open backlog causes on every `main` merge), instead of two:

- **#870** — cargo `non-security-minor-patch` group (11 updates, `Cargo.lock`)
- **#871** — github-actions `actions-minor-patch` group (5 action bumps across 25 workflows)

Both are **patch/minor within-major** (no breaking bumps). The 6 open **major** groups (#891 npm, #873 actions, #872 cargo, #801 docs, #799 TS 6→7, #795 held-majors) are deliberately kept **separate** — majors break and need individual verification/owner sign-off.

## Merge plan

Scheduled to **merge overnight** (Mac idle) once CI is green — so the downstream rebase of the remaining PRs happens off-peak.

Supersedes #870 and #871 (close them on merge).


gmhelmold:
Closing this combined batch — combining backfired. Diagnosis from its CI:
- The **cargo** minor/patch group (#870) actually FIXES 3 advisories (RUSTSEC-2025-0119, RUSTSEC-2025-0134, RUSTSEC-2026-0176 — cargo-deny now reports them as unused `ignore` entries in deny.toml), which is good, BUT it also fails the **canonical-vectors PR gate** — a real behavior signal that needs root-causing before merge (a minor dep bump shouldn't move a canonical test vector).
- The **actions** minor/patch group (#871) has only stale CI + the known cargo-fuzz Mac flake, but needs a fresh run.

Redoing dependabot as a focused pass: (1) merge the safe actions group after a re-run, (2) root-cause the cargo canonical-vectors break + prune the now-unused deny.toml ignores, (3) triage the 6 major groups individually for owner sign-off.

PR #906, opened 2026-07-21
ci(deps): bump the actions-major group (setup-* v7, slack-github-action v4)

gmhelmold:
Recreates #873 on green `main` (cherry-picked; dependabot authorship + sign-off preserved). The actions `major` group:

- `actions/setup-node` 6→7, `actions/setup-python` 6→7, `actions/setup-go` 6→7 — **safe**: every workflow pins an explicit `node/python/go-version` (22 / 3.12 / 1.22), and the setup-* v7 major only bumps the *action's own runtime* (drops old Node), not the toolchain it installs.
- `slackapi/slack-github-action` 3→4 — used only in `terraform-drift.yml` (a scheduled drift-notification, non-PR-gating) via the modern `method: chat.postMessage` + `token` + `payload` Bot-token technique, which v4 retains. Low blast radius.

Supersedes #873.


gmhelmold:
Deferring: the setup-* v7 majors break two OpenAPI meta gates (both GREEN on main, so it's the v7 bumps, not pre-existing):
- **breaking-change (oasdiff):** `go install oasdiff@v1.10.27` → `version constraints conflict` — setup-go v7 installs a newer Go toolchain that oasdiff's go.mod can't satisfy.
- **redocly lint:** `Property 'nullable' is not expected here` (openapi/corelink-v1.yaml:2119) — setup-node v7 pulls a newer redocly that validates as OpenAPI 3.1, flagging the spec's 3.0-style `nullable`.

Adopting needs a downstream-tooling fix (pin oasdiff/Go, and either pin redocly or migrate the spec off `nullable`) — a separate focused effort. The setup-* v7 majors are held until then.

PR #927, opened 2026-08-01
fix(worker): durable container idle reaper — a once-started container was immortal and billed 24/7

gmhelmold:
## The bug

`IDLE_TIMEOUT_MS` (30 min) was enforced by an **in-memory `setTimeout`** armed on the request hot path, while the health check ran on a **durable storage alarm** re-armed every 30 s. Those two lifetimes are not the same:

- The moment the DO's isolate is evicted — **every deploy, every recycle** — the `setTimeout` evaporates.
- The alarm chain is rehydrated from storage and keeps probing the container **forever**.
- The alarm path never re-armed the idle timer.

So **any container that started once was health-checked, and billed, in perpetuity** — keeping both the container AND the Durable Object awake. The platform's own `container.setInactivityTimeout` was never called either: it exists only in a doc-comment (a `designed-vs-wired` gap).

## Evidence (Cloudflare invoice IN-72804238, Jun 25 – Jul 24)

| Line item | Qty | Amount |
|---|---|---|
| Container Memory (GiB-s) | 33,724,894 | **$84.09** |
| Durable Objects Compute Duration | 599,553 GB·s | **$12.50** |
| Container vCPU | 286,536 | $5.28 |
| Container Disk | 71,030,170 | $4.92 |
| Workers requests / R2 / D1 / KV | 0 | **$0.00** |

33.7M GiB-seconds ≈ **13 GiB resident around the clock for a full month**, against **zero** billable traffic. The bill was idle burn, not usage — ~84% of a $111.79 invoice.

## The fix

The reaper moves **into `alarm()`** (durable) and compares a new persisted `LifecycleState.lastActivityMs` against `IDLE_TIMEOUT_MS`. On expiry it emits the existing `container_died` telemetry (audit-before-mutation preserved), destroys the container, and **stops re-arming the alarm** so the DO hibernates too — the second half of the leak.

`lastActivityMs` is touched in-memory on every proxied request (zero hot-path storage cost) and persisted by the alarm's existing lifecycle write, so it is at most one health-check interval stale — negligible against a 30-minute window.

Three correctness details, each one a way this could have gone wrong:
1. An **absent** `lastActivityMs` (state persisted before the field existed) backfills the clock instead of reaping — an upgrade never kills a live container mid-request.
2. The **double-fire dedup** path now re-arms the alarm instead of silently returning; a bare early-return there would orphan a running container with neither health checks nor reaper.
3. **`degraded`-but-running** containers stay IN the chain (probe skipped, semantics preserved) — dropping them would leave exactly one immortality path open.

## Verification

- `worker/tests/durable_object.test.ts`: 7 new cases — reap-on-expiry + no-reschedule, health-check-and-reschedule while active, pre-fix backfill, a fresh request beating stale persisted activity, dedup keeps the chain, and both degraded arms. **5 of the 7 fail against the pre-fix source** (verified by stashing the fix).
- Worker suite **585/585**.
- Typecheck error count **unchanged** (14 pre-existing, none in `durable_object.ts`).
- OKF reconciled (`planes/durable-object`, `planes/request-flow`): `validate_okf` 160 concepts / 0 stale / 0 drift; `validate_specs` 468/11.

Worker-only; ships on the next `corelink-api` deploy. Expected effect: container-memory COGS drops from always-on to proportional-to-activity (recently-active tenants + a bounded 30-min tail).


gmhelmold:
## Cold review applied — 3 more immortality/availability doors closed

An independent adversarial review confirmed the diagnosis (the `setTimeout`/alarm lifetime mismatch is real; nothing else reaped — `wrangler.toml` sets no inactivity knob and `setInactivityTimeout` is called nowhere) and returned **FIX-FIRST** with two HIGH findings. Both were legitimate and are now fixed in this PR, plus a MED:

**(a) HIGH — `startContainer`'s failure arms leaked resident containers.** Both the startup-health-check-failed and the `container.start` threw arms only re-labelled lifecycle `"stopped"` and returned. But `container.start()` had already taken effect, so a container that came up and then wedged on `/_health` (the R2/S3 init alone is measured at 26 s+) was left **resident, billed, and with no alarm chain to reap it** — the same immortality this PR exists to close, entered through the failure door. This is very likely the mechanism behind the per-DO `container_start_threw` wedge that previously cleared only on an image roll. Both arms now `destroyContainer()`.

**(b) HIGH — the alarm chain had no repair path.** `alarm()` owned both the health chain and the reaper but had no `try/finally`: any throw before its tail `setAlarm` (a `storage.put` fault, the PagerDuty POST, `hashForLog`) would end the chain permanently once CF exhausted its bounded retries — immortality by a third door, since `ensureContainerRunning` early-returns `ok` for a running container without arming anything. `alarm()` is now a thin `try/finally` shell around `alarmTick()` that **always** re-arms unless the tick reports a deliberate end — the posture `ReplicationCoordinatorDO.alarm()` already used — and the DO `fetch` path self-heals an already-lost chain with a cheap `getAlarm() === null` check.

**(c) MED — the reaper could kill an in-flight request.** It evaluated the idle predicate, then `await`ed `hashForLog` + the outbound `emitLifecycleEvent` POST (seconds-scale) before destroying — yield points where a queued `fetch()` passes `ensureContainerRunning` and starts proxying, only to be killed mid-flight. The predicate is now **re-checked immediately before `destroyContainer`** and the reap aborts if a live request raced in. (Pre-existing shape, but it only becomes reachable now that the reaper actually fires.)

Also fixed: the `degraded` and dedup arms wrote no lifecycle at all, so a degraded container's idle clock could revert to an arbitrarily old value on eviction — both now persist.

**Text corrections from the review:** the CHANGELOG's "~84% of the invoice" was the dollar figure mistaken for a share — it is **79%** ($84.09 of $106.79 in usage lines); the "persisted by the alarm's existing lifecycle write" claim was true only on the healthy arm and is now accurate; and the OKF citations (`CONTAINER_PORT` line, the "every non-reaping arm re-arms" sentence, the range boundaries) are corrected.

**Verification after the fixes:** worker suite **591/591**; the reaper suite is now **13 cases, 8 of which fail against the pre-fix source**; typecheck error count unchanged (14 pre-existing, 0 in `durable_object.ts`); `validate_okf` 160 concepts / 0 stale / 0 drift.

The review's remaining findings are deliberate follow-ups, not merge blockers: a single request longer than the 30-min idle window (streaming export / very slow large upload) can still be reaped mid-stream — there is no in-flight counter, a pre-existing class — and cold starts become real again for every DO, so a p99 first-request latency increase is the expected, intended trade (that is what "COGS proportional to activity" costs).

PR #931, opened 2026-08-01
fix(ci): warm routes before critical-flows instead of racing the dev compiler

gmhelmold:
## The bug

The suite runs against `next dev`, where Next compiles a route **on first request**. On a 2-core hosted runner that first-compile lands **inside** a spec's 30 s locator timeout — so failures hit innocent code, and whichever route was requested first won the race.

That is why this read as "flaky" rather than "misconfigured": it blocked #923 **four times on 2026-08-01, on a different spec each run** (`h1#audit-heading`, `h1#op-heading`, `h1#page-heading`), while the same specs passed locally in 16–57 s and firefox/webkit finished in 2–3 min against chromium's 10–20.

## The obvious fix is WRONG — tried, verified red, reverted

`next build` + `next start` was the first attempt. It cannot work: the suite's synthetic-session auth bypass is **double-gated** in `src/middleware.ts`:

```js
const isE2E = process.env["NEXT_PUBLIC_E2E_TEST_MODE"] === "1" &&
              process.env["NODE_ENV"] !== "production";
```

`next start` sets `NODE_ENV=production` → the bypass switches off → **every** spec fails on the 403 panel. Verified: all three browsers red on that approach.

That second gate exists so a stray `NEXT_PUBLIC_E2E_TEST_MODE=1` can never disable auth in a production deploy. **It was not weakened to make a test suite pass** — trading a flaky gate for an auth hole is not a fix.

## The actual fix

A Playwright `globalSetup` (`apps/admin-ui/tests/e2e/warm-routes.ts`) fetches each of the 11 routes the specs visit — **after** `webServer` is up, **before** the first assertion — so the compile is paid off the assertion clock.

- **Sequential on purpose:** parallel cold compiles contend for the same 2 cores, which is the exact failure being designed out.
- **Never fails the run itself:** a warm miss is logged and the spec that owns the route reports the real, readable failure. Warming is an optimisation, not a gate.
- **No workflow change, no production-server change, security guard untouched.**

## Verification (clean room — the first attempt's "pass" was not one)

Full disclosure: my initial local verification of the `next start` approach was **invalid** — leftover dev servers were listening on port 3010 and the suite passed against those, not against the production server I believed I was testing. CI then failed on all three browsers. Re-verified properly:

- port confirmed free (`lsof -ti:3010` → 0), `.next` deleted
- `[warm] 11 routes compiled in 14103ms`
- **13/13 specs green in 48.4 s**

⚠️ **Matrix note:** the CI-cost-diet change on `main` made the browser fan-out event-scoped — a `pull_request` runs **chromium only**. I am dispatching the full 3-engine matrix on this branch and will report it here before merging.


gmhelmold:
## ⛔ CI is blocked org-wide by GitHub Actions billing — not this PR

The dispatched 3-engine matrix failed on all three browsers in **~2 seconds with ZERO steps executed** and `runner: ""` (no runner ever allocated). The check-run annotation is definitive:

> "The job was not started because recent account payments have failed or your spending limit needs to be increased. Please check the 'Billing & plans' section in your settings"

This blocks **every hosted workflow in the repo** (PR gates AND `cf-deploy-prod` / `container-build-push`), not just this PR. It is the same block that hit on 2026-06-25/26/28 — first recurrence under the migrated `HuGR-Labs` org.

**Owner-only fix:** GitHub → org **Settings → Billing & plans** (clear the failed payment / raise the spending limit).

### Where this PR stands

The fix is **verified locally in a clean room** — port confirmed free (`lsof -ti:3010` → 0) and `.next` deleted, so no stale server could flatter the result:

```
[warm] 200 /sign-in
… 11 routes …
[warm] 11 routes compiled in 14103ms
13 passed (48.4s)
```

Against 10–20 min and red in CI before. What is still **unverified** is the firefox/webkit matrix, which needs hosted CI. I will not merge on a local-only signal for a CI-infrastructure change — this PR waits for billing.

*(Diagnostic note for next time: when a job fails in ~2 s the log is absent, so `gh run view --log` returns "log not found" and looks like a tooling error. `gh api repos/<org>/<repo>/check-runs/<job_id>/annotations` is the fast, definitive read.)*

PR #918, opened 2026-08-01
chore(ci): CI cost diet — remove ~7,900 measured hosted Actions min/month, zero gates removed

gmhelmold:
## Why

The repo moved to **`HuGR-Labs`** on a **Free plan = 2,000 hosted Actions min/month**, while
the owner's June 2026 billing read was **13,758 min**. Paid minutes, not the Mac fleet, are
the scarce resource.

**No quality gate is deleted.** Every change is one of: move a *light* job to the idle
self-hosted fleet, reduce the *cadence* of a schedule, tighten a `paths:` filter, or cancel a
redundant run.

> **This PR was reproved by an independent cold review and has been corrected.** The central
> justification of the first revision was false, the arithmetic did not hold, and four of the
> proposed cuts were unsafe. All seven findings were re-verified against live Actions data and
> acted on. See **"What the review changed"** at the bottom.

## Method (corrected)

Savings are measured against **July 2026** — the current configuration — job-by-job from
`/actions/runs/{id}/jobs`, applying GitHub's **round-up-to-the-minute per job** rule and keying
each job on its `labels` (hosted vs self-hosted).

**June is deliberately NOT the denominator**, and no "% of the June bill" is claimed, because
June minutes cannot be projected forward for two structural reasons:

1. The `pull_request`/`push` triggers on `codeql`, `coverage`, `cas_foundation`,
   `ffi-matrix-ci` and `s10-ship-gate` were **deleted on 2026-06-04**. Those are dead minutes
   that cannot recur — `codeql` ran **90** times in June but only **29** on `schedule`;
   `coverage` 88/28; `cas_foundation` 91/27; `ffi-matrix-ci` 67/27; `s10` 68/27.
   Cadence changes only act on the `schedule` slice.
2. Until `2c5145e4` (2026-07-12, *"migrate 16 chronic-red nightly gates off the broken
   self-hosted Mac"*) those same heavy gates ran on the **free** fleet. Their June *hosted*
   cost was near zero; their cost today is not. Verified: a June `coverage` scheduled run's
   job carries `labels: [self-hosted, mac, corelink-builder]`.
3. `cas-canary.yml` (created 2026-07-10) and `rustfmt.yml` (2026-07-20) had **zero** June runs
   and cannot appear in a June-denominated figure at all.

## Estimated saving — measured, July basis

| lever | min/month |
|---|---:|
| Runner: 21 sub-minute jobs → fleet | **~2,680** |
| Frequency: daily → weekly (schedule slice only) | **~4,310** |
[...]

PR #932, opened 2026-08-02
ci: inventory the fabric box, then move the first PR gate onto it

gmhelmold:
## Why

The repo is **private**, so every `ubuntu-latest` minute is billed against a **2,000/month** quota — **1,671 of which were already spent on 2026-08-01**. The cost diet ([#918](https://github.com/HuGR-Labs/corelink-server/pull/918)) cut real waste but the remaining spend is still several times the quota. **Cutting does not close the gap; moving the work does.** A self-hosted job bills zero.

## What this does

**1. Inventories the fabric box instead of guessing from the Dockerfile.**

Reading `deploy/runner/Dockerfile` answers "what was installed", not "what the box has" — and it gets two decisive things wrong. The measured result (`probe-corelink-box`, run `30724931256`):

```
box     4 vCPU · 11.9 GiB RAM · 16 GB free · Ubuntu 24.04.4 · Firecracker · sudo
PRESENT rustc/cargo 1.91.1  ← the exact pin in rust-toolchain.toml
        clippy · rustfmt · cargo-deny 0.19.8 · cargo-audit 0.22.2
        git · curl · jq · clw 0.1.4
        bundled node20/node24 under externals/ (this is why JS actions work)
ABSENT  python3 · node/npm/pnpm on PATH · docker · gh
```

Two findings the Dockerfile alone would have gotten wrong:

- **JS actions are not blocked.** The runner ships its own Node, which is why `actions/checkout` ran fine on the box in #928 despite no `node` on `PATH`. Only steps that *shell out* to npm/pnpm are affected.
- **rustup honours `rust-toolchain.toml`.** The image declares 1.96.0; the box reports **1.91.1** — it synced to the repo's pin on first use. So Rust gates need no toolchain work at all.

The probe stays in the repo as a `workflow_dispatch` tool, to re-run whenever the runner image changes.

**2. Moves the `rustfmt` gate to `runs-on: corelink`.** One line. No step touched.

## Classification of all 71 hosted workflows

Derived by parsing every `run:` block for absent binaries and cross-checking for a setup action that provides them:

| | count | |
|---|---:|---|
| **A** — moves today, box has everything | **25** | `rustfmt`, `cargo-deny`, `coverage`, `corelink-{server,worker,meta,hash,reapi}`, `trivy`, `gitleaks`, `mutation-pr`, … |
| **B** — unlocks with `apt-get install python3` in the image | **13** | `cargo-audit`, `okf_wiki`, `region_pinning`, `perf-regression`, `cas_foundation`, … |
| **C** — moves, but pays a toolchain download per run | 17 | carry their own `setup-node`/`setup-python` |
| **D** — needs more than python3 in the image | 15 | `pnpm` / `gh` |
| **E** — must stay hosted | 1 | `smoke-install` (docker) |

This PR does **A: 1 of 25** deliberately — one real PR-blocking gate, so the fabric is proven under actual gate traffic before 24 more files change.

## Why no step was trimmed

On a self-hosted runner the billed minutes are zero regardless of how many steps run. Removing the now-redundant `dtolnay/rust-toolchain` step would save nothing and change what is tested. **Minimum diff, maximum saving.**

## Revert

One line per workflow, back to `ubuntu-latest`.


gmhelmold:
## Closing this — the measurements are salvaged, the code was overtaken

**Verdict: close unmerged.** Not because the work was wrong. The analysis in
here was the most rigorous thing anyone had done about the fabric box, and it
was right on the two points that mattered most. The tree simply moved out from
under it over the following day.

### Signal ratio

Of **1148 added lines, 35 are the actual change** (`runs-on: ubuntu-latest` →
`runs-on: corelink`, 35 job-level edits across 22 workflow files). The rest:

| | lines | |
|---|---:|---|
| accidentally-committed build output | **698** | `apps/docs/playwright-report/a11y-results.json`, `.../a11y/index.html`, `apps/docs/test-results/.last-run.json` |
| agent telemetry | **11** | `.maestro/*.jsonl` |
| one identical 9-line comment block, pasted **34×** | **333** | |

**~3% signal** — and enough volume to earn the `size:XL` bot comment above,
telling the author to split what is genuinely a one-line-per-file change. That
was the tooling's fault, not yours: `.gitignore` covered none of those paths,
and neither did `apps/docs/.gitignore`. **Fixed in #1019** so the next PR can't
inherit it.

### What superseded each half

| half of this PR | superseded by |
|---|---|
| the probe workflow (`probe-corelink-box.yml`) | **`.github/workflows/runner-probe.yml`**, already on `main` — a strict superset (it also measures rustup toolchain resolution cost, cargo-home seeding, and cold-compile timings) |
| the migration wave | **#1003** (wave 1) + **#1005** (wave 2), both merged. `main` now has `runs-on: corelink` in 11 files, **6 of which this PR also patches** — exactly its 6 conflicting files / 9 conflict hunks, every one resolving as *take-main* |

### Two premises the tree overturned

- This PR argued for **keeping `dtolnay/rust-toolchain`**. `main` has since
  replaced it with `bash scripts/ci-assert-pinned-toolchain.sh`.
- The blanket **"no step is changed"** policy — sound about *billing*, but it
  would have walked straight into the shared-`$HOME` `os error 2` defect that
  **#980** later had to fix with `if: runner.environment == 'github-hosted'` on
  every `Swatinem/rust-cache` step. Migrating a workflow means reviewing its
  steps, not just its label. That lesson is now written down.

### What is salvaged, and where it lives

Both artifacts now live in **`docs/internal/ci-runner-fabric-box.md`** (#1019).
Nothing in `docs/` had recorded any of this — `git grep 'runs-on: corelink' -- docs/`
returned nothing while 11 workflows on `main` already used the label.

1. **The measured box inventory**, re-verified against the probe run this PR
   cites ([`30724931256`](https://github.com/HuGR-Labs/corelink-server/actions/runs/30724931256))
   plus the newer `runner-probe.yml` run
   ([`30825772599`](https://github.com/HuGR-Labs/corelink-server/actions/runs/30825772599)) —
   including the finding this PR deserves the credit for and which had already
   misled people twice: **`rustup` honours `rust-toolchain.toml` over the
   image's default** (image ships 1.96.0; the repo resolves 1.91.1), and
   **bundled `externals/node20`/`node24` mean JS *Actions* run even though
   `node` is not on `PATH`** — while a `run:` step calling `node`/`pnpm` does
   not. Both were correct here and both are preserved.
2. **The A–E classification**, recorded as a planning table alongside the
   post-dating cost data.

### Two corrections found while verifying, for the record

- **`coverage` is not "class A, moves today."** It needs `llvm-tools-preview`;
  the probe counts **0** on the box. This PR was only right about it because
  the no-step-changes policy kept the `dtolnay` step that installs the
  component — and that step is now gone from `main`.
- **`go` was never probed** by run `30724931256`. The later probe confirms it
  absent, so the claim is now actually sourced.

### The thing that changes the priority order

Measured after this PR was written (2026-08-01→03): **4429 billed min ≈
$26.57**, and the **four heaviest are 66% of the bill** — `admin-ui-e2e`,
`admin-ui-ci`, `docs-ci`, `lighthouse-ci` — all `ubuntu-latest`, all **class
D**, all blocked on `node`/`pnpm`. The entire Rust class-A opportunity this PR
targets is **~$39/mo, about a quarter** of what node+pnpm in the image unlocks.

**So the next high-leverage move is a one-file image change, not another
label-swap wave.** That is a conclusion this PR's own inventory made possible —
it just wasn't visible from where it was standing.

Also now recorded: **`perf-regression` must not be moved** (class B on paper,
but criterion baselines are CPU-bound — changing the runner invalidates the
gate itself).

Branch `ci/fabric-capability-probe` is **kept**, not deleted.

Thanks for measuring instead of guessing. That part outlived the code.

PR #933, opened 2026-08-02
fix(ci): move macOS-safe PR gates to the self-hosted mac fleet (hosted runners billing-blocked)

gmhelmold:
## Why now

From **~2026-08-02 00:40 UTC** every `ubuntu-latest` job fails **before executing a single step** — ~2 s, zero steps, `runner: ""` (no runner ever allocated). The check-run annotation is definitive:

> "The job was not started because recent account payments have failed or your spending limit needs to be increased. Please check the 'Billing & plans' section in your settings"

Same block as 2026-06-25/26/28 — **first recurrence under the migrated `HuGR-Labs` org**. Clearing it is owner-only (org Settings → Billing & plans), and while it lasts it kills PR gates **and** `cf-deploy-prod` / `container-build-push`.

## What this does — explicitly TEMPORARY

Moves the PR-blocking gates that are **macOS-safe** to `[self-hosted, mac, corelink-builder]`:

`admin-ui-ci`, `admin-ui-e2e` (both jobs), `worker-vitest`, `signup-worker-vitest`, `okf_wiki`, `lighthouse-ci`, `cargo-audit` (both jobs).

Each moved job carries an inline banner naming the reason **and the revert condition** (billing cleared, or the corelink runner fleet GA) so this cannot silently become permanent.

`playwright install --with-deps` also drops `--with-deps`: it shells out to **apt** and hard-fails on macOS.

## Deliberately NOT moved

`trivy`, `semgrep`, `cosign-sign`, `cf-deploy-prod`, `ffi-matrix-ci`, `cas_foundation`, `smoke-install`, `backup-daily*` — these pin a **Linux binary/toolchain**, so moving them means editing a pinned checksum. That is a security-relevant edit and not one to make at speed to chase a green tick. They stay blocked until billing clears.

## Fleet sanity-checked before editing

- 4 mac builders **online, 3 idle**
- **69 GB free disk** — the "~98 % full" note in the ops memory is **stale**
- node 22.17.1, pnpm 10.32.1, warm chromium playwright cache (firefox/webkit download on first use)

## Verification

This PR verifies itself: if the moved gates come back to life on this PR, the migration works — that is the acceptance test, and it is visible in the checks below. YAML parse + `runs-on` values confirmed for all 7 files before pushing.


gmhelmold:
Closing — the trigger is gone and the destination is the wrong one.

**Why the trigger is gone.** The org's Actions billing block cleared. Evidence, not
assumption: an `ubuntu-latest` job completed `success` at **2026-08-02T01:15:37Z**,
after the block window. The billing failure mode is unmistakable (a ~2-10 s job,
zero steps, `runner: ""`, and the "recent account payments have failed" annotation),
so a hosted `success` is proof the quota is honoured again.

**Why the Mac is the wrong destination even when it works.** `CLAUDE.md` is explicit:
the self-hosted mac fleet **is the founder's own machine**, heavy CI bursts have
already crashed it once, and local builds are capped at `CARGO_BUILD_JOBS=4` for
exactly that reason. This PR would put `admin-ui-e2e` (two jobs), `lighthouse-ci`
and `okf_wiki` on it as *PR-blocking* gates — i.e. every push, forever, since a
"temporary" runner label with a green tick has no forcing function to come back off.
The 69 GB-free / 3-idle snapshot in the description is true and still doesn't make
this the right home for recurring gate load.

**Where this work actually goes.** #932 moves gates onto the **CoreLink fabric**
(`runs-on: corelink`) — dedicated ephemeral boxes, zero billed minutes, zero load on
the founder's hardware, and it dogfoods the product we sell. 21 workflows / 34 jobs
are already green there. The overlap with this PR is bucket D (the pnpm/node gates);
those are being unblocked on the fabric rather than parked on the Mac.

**Kept, not discarded.** Two findings here are real and are carried forward:

1. `playwright install --with-deps` shells out to `apt` and hard-fails on macOS. That
   is a genuine portability bug in the step and it is worth fixing on its own terms.
2. The runner-label inventory in the description (node 22.17.1, pnpm 10.32.1, warm
   chromium cache) is useful capability data and is folded into the fabric bucket work.

The branch is left in place; nothing is deleted. Reopen if billing regresses before
the fabric covers these gates.

gmhelmold:
## Scope note: this PR stays at 7 workflows — #932 owns the corelink/fabric side

Billing is cleared, but the standing instruction is **no GitHub-hosted runners** (cost). I started to move the remaining ~50 `ubuntu-latest` jobs onto the `corelink` Linux fleet — Linux→Linux is a drop-in, so every pinned `linux_x64` binary + sha256 (trivy, gitleaks, semgrep, cosign, oasdiff) keeps working untouched, which is exactly why those could never go to the mac fleet.

**I reverted that and did NOT push it**, because #932 ("inventory the fabric box, then move the first PR gate onto it") is already doing that migration — and doing it the right way, one gate at a time after inventorying the box, rather than a 50-file blanket rewrite whose tooling assumptions are unverified. A blanket push would have collided head-on with its 28 files.

**Verified no file overlap** between this PR's 7 and #932's 28. The split is clean and complementary:
- **this PR → mac fleet:** the JS/TS + Node gates (`admin-ui-ci`, `admin-ui-e2e`, `worker-vitest`, `signup-worker-vitest`, `okf_wiki`, `lighthouse-ci`, `cargo-audit`)
- **#932 → corelink/fabric fleet:** the Rust crate CIs + `cargo-deny` + the byok drills

The Linux-pinned security scanners (`trivy`, `gitleaks`, `semgrep`, `cosign-sign`) and `cf-deploy-prod` still need a home on the fabric box — that belongs to #932's incremental sweep, not here.

### Evidence this fleet move works

On this PR, running on the mac fleet: **`playwright legacy (chromium)` PASS**, **`cargo-audit PR gate` PASS (4m17s)**, `actionlint` PASS, `dco` / `secrets-matrix drift` / `Verify all GHA actions SHA-pinned` PASS. `playwright legacy` is the meaningful one — it proves Playwright runs on macOS after dropping `--with-deps` (which shells out to apt and hard-fails there).

`typecheck` shows as failed but the API reports **`cancelled` with zero failed steps** — queue contention on 4 macs, not a regression.

PR #934, opened 2026-08-02
fix(ci): weekly (not daily) for the 8 crons that re-prove an unchanged tree — and 6 of them have never passed

gmhelmold:
## Measured, not guessed

3 days of runs: **1000 runs / 3851 wall-min across 69 workflows**, and **40 daily crons**.

The waste concentrates in checks whose **only input is the tree itself**:

| cron | was | cost |
|---|---|---|
| `tla_check`, `tla_billing_check`, `tla_dsr_erasure_check`, `tla_region_residency_check`, `tla_runbooks_check` | daily 07:00→08:20 | 5 model-checkers/day against `specs/tla/*.tla` files that change on the order of **months** |
| `region_pinning` | daily | **~33 min per run** |
| `reproducible-build` | daily | proves determinism of an unchanged tree |
| `proptest-density-gate` | daily | lint-shaped, over unchanged sources |

## The naive cut would have been a regression

**6 of the 8 have NO `pull_request` trigger** — the cron is their *only* gate. Dropping them to weekly on its own stretches worst-case detection from 24 h to **7 days**. That is trading a real safety property for minutes, which is not a trade worth making.

So each one **gains a path-gated `pull_request` trigger AND drops to weekly**:
- TLA+ set → `specs/tla/**` + its own workflow file
- `reproducible-build` → `Cargo.lock`, `Cargo.toml`, `crates/**`, `Dockerfile`

**Net effect — strictly better than today:** a real change is caught **immediately on the PR** (today: up to 24 h late), while an unchanged tree is re-proved **1× instead of 7× a week**.

## The rule applied

> Stay **daily** only if the **SIGNAL** changes daily.

- **new advisories** → `cargo-audit`, `pnpm-audit`, `semgrep` — kept daily
- **production state** → the e2e canaries, backups, billing reconcile — kept daily
- **randomness** → `fuzz-nightly` — kept daily
- **only the code could change it** → the 8 above — weekly, because the PR gate now covers the change

Everything left daily was left there on purpose, not by omission.

Weekdays are staggered Mon–Fri and each cron keeps its original minute/hour, so there is no thundering herd and the low-traffic window is preserved.

Complements #918 (earlier cost diet) and #933 (fleet move off GitHub-hosted runners). No gate is deleted.


gmhelmold:
## Course correction after measuring

This PR originally added a path-gated `pull_request` trigger to the 6 crons that had none, arguing coverage would go **up**. I measured their actual pass rate before merging, and it inverted the plan:

| workflow | successes / runs |
|---|---|
| `tla_check` | **0 / last 100** (back to 2026-06-13) |
| `tla_dsr_erasure_check` | **0 / 25** |
| `reproducible-build` | **0 / 25** |
| `region_pinning` | 2 / 25 |
| `tla_billing_check` | 7 / 25 |
| `tla_runbooks_check` | 7 / 25 |
| `tla_region_residency_check` | 6 / 25 |
| `proptest-density-gate` | **23 / 25** ← the only healthy one |

The TLA+ set shares one root cause — a **supply-chain pin violation**, not a spec failure:

```
TLC SHA-256 supply-chain pin violation (ADR-0042 §A1)
actual:   e22f8ffb4bacdea0a871f444dd94fe5fb0d8013b3388ae39e82e26f852c735d5
expected: 33de7da9ce1b7fffb9d1c184021178dbb051747be48504e65c584c423721a32e
Re-pin requires Security review + ADR-0042 §A1 update
```

The install step hard-fails before a single spec is ever checked. Attaching a PR trigger to a gate in that state would block **every** pull request while proving nothing — so the triggers are removed, and each workflow now records its measured pass rate inline plus the condition for re-adding it.

**I deliberately did not re-pin TLC.** The workflow's own error says it needs Security review + an ADR-0042 §A1 update — that's an owner call on a supply-chain artifact, not something to slip into a CI cleanup.

The daily→weekly change stands for all 8: that's the cost saving, and it's strictly better for gates that cannot pass anyway.

Also merged `origin/main` — this branch predated #926/#930, so its diff would otherwise have **reverted** the route-matcher basePath fix.

PR #935, opened 2026-08-02
fix(container): the free signup row counted as a subscription — nobody could buy

gmhelmold:
## The defect

`POST /v1/onboarding/tier-select` answered **`409 already_active` to the first purchase attempt of every account that had ever signed up.**

Measured in prod D1 on 2026-08-02:

| tier | subscription_state | rows |
|---|---|---|
| **free** | **active** | **117** |
| pro | pending_checkout | 3 |
| enterprise / max / pro / solo / starter | active | 1 each |

117 registered tenants, every one structurally unable to pay.

## Root cause

Signup seeds every tenant with `tier_selections('free','active')` — deliberately, so billing reads `active` rather than `inactive` (`seedTenantEntitlements`, mirrored by `githugr_provision.ts`). But `has_active_subscription` was:

```sql
SELECT 1 FROM tier_selections WHERE tenant_id = ?1 AND subscription_state = 'active'
```

No tier predicate. So the seed row answered *"this tenant already has an active subscription"* and the guard rejected the upgrade before Stripe was ever reached.

**The implementation contradicted its own contract on both sides.** The guard's comment says it refuses "a *second* subscription of the SAME kind"; the OKF concept `launch/tier-model` calls Free an instant *activation*, not a subscription. Only the SQL disagreed.

## The fix

`AND tier != 'free'`, extracted to a named constant so the predicate is testable without live D1.

The one-row-per-tenant shape is untouched: `persist_pending_checkout` upserts `ON CONFLICT(tenant_id)`, so an upgrade flips the **same** row `free/active → pending_checkout → paid/active`. The UNIQUE partial index `idx_tenant_active_subscription` still holds.

### Second-order hole closed in the same pass

The signup-worker's activation UPSERT was guarded `WHERE subscription_state <> 'active'` — documented as the fallback for a lost container `pending_checkout` persist. But a free tenant's row **is** `active`, so that fallback silently did nothing for exactly the tenants it existed to rescue: Stripe takes the money, tenant stays on free. Now `… <> 'active' OR tier = 'free'` — an active **paid** row is still protected from a redelivered/out-of-order webhook; a free row is not.

The test-only `corelink-tier-selection` ledger (verified **not** wired in prod — instantiated only by its own tests and benches) mirrored the same semantics in **both** guards, including `on_checkout_completed`, where it would have rejected the very activation the customer paid for. Both now call the new pure predicate `TierSelectionRow::holds_paid_subscription`.

## Tests — proven against the pre-fix code, not merely written

- **3 cases** in `mutation_kills.rs`: free→paid reaches Checkout; a paid active row **still** refuses a second paid subscription; the webhook activates an upgrading free tenant. **2 of the 3 fail** when the `tier != Free` half of the predicate is reverted (verified by reverting it).
- **An always-on SQL-predicate guard** in `tier_select_store.rs`. The behavioural D1 proof is `#[ignore]` (needs a live database), so CI would otherwise not catch the predicate simply being deleted.
- **A signup-worker vitest case** that fails against the shipped guard (verified by reverting it).
- The pre-existing live-D1 test `d1_persist_free_active_then_active_true` had **frozen the defect** — it asserted that a free row counts as an active subscription — and is corrected and renamed.

## Verification

| gate | result |
|---|---|
| `corelink-tier-selection` | 52/52 |
| container `routes::tier_select` | 41/41 + 3 ignored (live-D1) |
| signup-worker vitest | 84/84 |
| `cargo clippy -D warnings` | clean |
| `validate_specs.py` | 479 validated |
| `validate_okf.py` | 160 concepts, 0 stale, 0 drift |
| `secrets-checklist-verify.sh` / `validate_secrets_matrix.py` | OK / `code_only=0` |

## Deploy note

The container fix ships on the **next container image roll** — a Worker deploy alone does not pick it up. The signup-worker half ships with the worker deploy.


gmhelmold:
## Predicate validated against live production data

The unit tests prove the logic; this proves the **blast radius**, by running both the old and the new predicate against the real prod `tier_selections` table (read-only):

```sql
SELECT (SELECT COUNT(*) FROM tier_selections WHERE subscription_state='active')                      AS old_guard_blocks,
       (SELECT COUNT(*) FROM tier_selections WHERE subscription_state='active' AND tier != 'free')   AS new_guard_blocks,
       (SELECT COUNT(*) FROM tier_selections WHERE subscription_state='active' AND tier  = 'free')   AS unblocked_by_fix
```

| | count |
|---|---|
| blocked by the **old** guard | **122** |
| blocked by the **new** guard | **5** |
| unblocked by this fix | **117** |

The 5 still blocked are exactly the genuinely-paid rows (`enterprise` / `max` / `pro` / `solo` / `starter`, one each) — they *should* keep being refused a second subscription on the cache axis, which is what the guard exists for. Nothing else changes: this is a read-side predicate, no data is migrated.

Note the fix ships on the next **container image roll** — a Worker deploy alone will not pick it up.

PR #946, opened 2026-08-02
fix(docs): the RBAC guides documented a CLI that has never existed

gmhelmold:
Fourth and largest pass of the phantom-surface sweep. Follows #942 (installer verb), #944
(installer arch + checksum), #943 (tutorial outputs).

## What was wrong

**28 phantom `corelink …` invocations across 12 files, in four languages** — `admin members
invite|list`, `admin pat mint`, `admin audit query|prove`, `pat revoke` — plus a phantom endpoint
`POST /v1/admin/pat/{id}/revoke` that exists nowhere in the repo.

**The fabrications were the larger half.** Deleted: invented `USER/EMAIL_HASH/ROLE/SINCE` and
`ACTOR_HASH/TARGET_HASH/DETAIL` tables no command ever printed; an invented JSON body for a
non-existent `GET /v1/admin/audit/events`; an invented `region="enam"`; and four CloudEvent type
names with **zero hits in the codebase**.

## What is there instead

The real surface shipped long ago — the guides were hiding it:

- **Dashboard:** `/customer/team`, `/customer/keys`
- **Dual-auth HTTP API** (PAT *or* Clerk session): `/v1/customer/team{,/invite,/{user_id}}`,
  `/v1/customer/keys{,/{pat_id}/revoke}`, `/v1/audit/analytics/*`, `/v1/audit/{tenant}/export`
- **Real CLI verbs:** `corelink audit export|verify|verify-ndjson|tail`

Every route documented against the route table. Responses described in prose, never sampled.

## Five more claims verified FALSE — none in scope when this started

1. **"A Clerk invitation email is sent."** There is no Clerk call on invite —
   `customer_d1.rs:1489`: *"NO synchronous Clerk call (OB-1) — the invitation EMAIL round-trip is
   an owner follow-up"*. The dependent "invites expire in 14 days" pitfall went with it.
2. **"A pending membership row is NOT yet written."** The `team_member` row **is** written at
   invite time with status `invited` (`customer_d1.rs:1531`) — which is exactly why the dashboard
   has a Pending-invitations list.
3. **`reason=CompromiseSuspected` on revoke.** The endpoint takes **no request body**
   (`routes/customer.rs:828`) and writes **no** `revocation_log` row — only
   `UPDATE pat SET revoked_at_ms`. The table is referenced zero times in the container crate.
4. **"Clients receive 401 on next request."** 401 is correct — but *"next request"* is not: the
   edge verify cache holds a positive result up to 5 s (`pat_verify_cache.ts:153`; ADR-0030 bounds
   revocation at 60 s p99).
5. **The role-grant rules.** `owner` is **never** grantable via invite (403 for everyone,
   `routes/customer.rs:936`) and only an owner may invite an `admin` (`:951`) — so the dropdown is
   not "grayed out" as the pitfall claimed.

## Method note, because it is the point

This ran as **four parallel agents against a frozen contract**, and **the contract I wrote carried
two of those falsehoods** — the audit-event names and the Clerk email. Two agents independently
flagged the first; one caught the second.

A verification claim *I* made — that `GET /v1/customer/team` leaks a raw `email` — was **also
wrong**: `routes/customer.rs:894` is the in-memory stub, and the D1 production path returns the
placeholder `"—"` (`customer_d1.rs:1471`), the raw address never being persisted.

An agent also **retracted its own** earlier correction: it had applied the audit crate's email-hash
helper to the team-member list, which uses a different one. The doc now asserts **no algorithm at
all** rather than a confident wrong one.

Every assertion in these guides is now backed by a read line, or absent.

## Validated against a measured baseline, not assumed

| | errors | warnings |
|---|---|---|
| `origin/main` | 796 | 19 |
| this branch | 796 | 20 |

`translation-quality-check --strict` is **chronically red on main** — the 796 are entirely
pre-existing. The single delta is one warning: a German banned-word regex matching the literal API
field name `email` in `{email, role}`.

Docusaurus build clean for **en** and **pt-BR**. That locale build is the gate that catches exactly
this class — it rejected a doubled `/docs/` link prefix earlier in this same sweep.

## Open, deliberately not fixed here

- **Product bug (admin-ui, not docs):** `TeamClient.tsx` tells the customer *"We've emailed them an
  invitation"* and *"They'll get an email invitation to join this tenant"* — but OB-1 means no email
  is sent. The docs are now silent rather than contradicting the UI. The invitee must sign up
  independently; the seat then binds by email hash. Filed separately.
- `corelink pat` / `gc` / `auth` in other docs, and Step 1's Clerk-SSO/MFA claims in the invite
  guide, remain unaudited.


gmhelmold:
Closing — this branch is the **whole 72-file wave**, not a slice. #950 turned out to be a strict subset of it (51 of the 72 files), so leaving both open means the same change reviewed twice and merged twice.

Split properly by deploy surface instead:

- **#950** — docs + marketing + scripts (the 51 files, including every RBAC guide this PR was originally about)
- **#952** — admin-ui customer surface (wire shapes + copy truth)
- **PR C, next** — activation telemetry (root worker + admin-ui bindings)

Nothing is lost: the branch `docs/rbac-real-surface` and `wave-checkpoint` both still hold the full tree. Splitting by deploy surface is deliberate — a worker problem should never force reverting customer-facing documentation.

PR #967, opened 2026-08-03
fix(signup): every free signup was granted paid Runners capacity

gmhelmold:
## The defect

`seedTenantEntitlements` wrote a `runners_entitlement('free', max_concurrency=1)` row on **every** Clerk `user.created`, and `worker/src/lib/githugr_provision.ts` mirrored it statement-for-statement on every githugr login.

That row is not a placeholder — **it IS the entitlement `handleRunnerMint` gate 5d checks**. Writing it grants the capacity. The comment justifying it ("so the runner cap gate reads a real entitlement") described the intent correctly and the effect wrongly.

It silently reverted **owner-ratified Option B** (Runners = a SEPARATE PAID axis from the cache tier, 2026-06-13) — a decision made precisely because the alternative *"leaked compute a cache-only tenant never bought"*. `migrations/d1/0070_runners_entitlement.sql` states the contract in its own header:

> the table starts empty … NO row → the tenant is cache-only and gets no runner cap … **Empty table = no cap = reject**

## Why the other gates did not compensate

| Gate | Source | Status for a free signup |
|---|---|---|
| 5a installation map | `github_install_callback.ts` | ✅ seeded on install |
| 5b offboarding | absent-by-default | ✅ passes |
| 5c repo allowlist | `github_install_callback.ts` | ✅ seeded on install |
| 5d entitlement | **this bug** | ✅ seeded on signup |

`github_install_callback.ts` contains **no entitlement check of any kind** — grepped: zero hits for `entitle`/`tier`/`paid`/`subscription` across its 387 lines. So the full chain was:

**free signup → install the public GitHub App → all four gates green → boxes spawn on our Cloudflare account.**

## Measured in prod D1 (`d64742ea`, 2026-08-02)

- **119 rows** `plan='free', max_concurrency=1`, created 2026-06-12 → **2026-08-01** (i.e. the live signup path was still writing them).
- 118 map to cache `tier_selections.tier='free'`; 1 to `pro`.
- **0 of the 119 hold a `runner_repo_allowlist` row** → the exposure was **latent, not active**.

That last number is the reason nothing was ever red: **no gate fires on a grant that has not yet been used.**

**Aggravating factor:** slot admission (`decideSlotAcquire`, corelink-runners `deploy/cloudflare/src/lib.ts`) is first-come-first-served against a GLOBAL `FLEET_MAX_CONCURRENCY = 20` with **no reservation for paying tenants** — so the free grant competed directly with revenue, not just with idle capacity.

## ⚠️ Root cause worth keeping

`max_concurrency INTEGER NOT NULL CHECK(max_concurrency > 0)` makes **"entitled to ZERO" inexpressible**. A constraint written to guarantee that a row always means a real entitlement is exactly what forced a writer wanting a harmless placeholder to grant real capacity instead. Absence IS the zero — that is the design, and it is now stated at both call sites so the next person does not re-derive the same mistake.

## Verification

Negative pins, each **proven RED against the restored bug** before being trusted:

| Pin | Failure against the bug |
|---|---|
| `webhook-e2e.test.ts` — `rowCount("runners_entitlement") === 0` | `expected 1 to be +0` |
| `githugr_provision.test.ts` — batch length 4 | `expected [...] to have a length of 4 but got 5` |
| `githugr_provision.test.ts` — H5 atomic batch | `expected 5 to be 4` |

- **846 tests green** after the fix (235 signup-worker + 611 worker).
- `tsc --noEmit` error counts **identical to main** (22 worker / 36 signup-worker) — all pre-existing, none in the touched files.
- `validate_specs.py` → 480 total, 0 failures.

## Not in this PR — deliberately

**The 119 existing prod rows are data, not code.** They are deleted as a separate operator step **after** this ships: cleaning first would let the still-live signup path re-seed them. Deploy order is `corelink-signup-worker` (manual `wrangler deploy` — it has no CI deploy path) + `corelink-api`, then the D1 cleanup.

Two sibling findings from the same audit are tracked separately and are **not** fixed here: the `max_vcpu_h` monthly ceiling is enforced only in fabricd's `ComputeGate`, whose route is unmounted in prod; and the webhook rate-limit (429) + mint-403 branches still strand a job with no dead-letter.


gmhelmold:
reopening to re-trigger pull_request workflows (none fired on open)

PR #980, opened 2026-08-03
fix(ci): gate cargo-cache steps to hosted runners — one runner's cache save was deleting ~/.cargo out from under the other four

gmhelmold:
## The failure

Four Rust jobs on #974 went red with the same error, on a PR whose diff was 14
workflow YAMLs + `CHANGELOG.md` + `CLAUDE.md` — **zero Rust files**:

```
error: could not parse/generate dep info at:
  …/target/release/deps/syn-e66560a1966a4c83.d
Caused by:
  No such file or directory (os error 2)
```

That `.d` line is the **symptom**. The cause is the line above it, in the same log:

```
error: could not compile `ctutils` (lib)
Caused by:
  could not execute process `rustc …` (never executed)
Caused by:
  No such file or directory (os error 2)
```

## Root cause

The self-hosted fleet is FIVE runners (`~/.gh-runners/runner-1` … `runner-5`) on
ONE Mac sharing ONE `$HOME`, therefore ONE `~/.cargo`.

`Swatinem/rust-cache` is built for GitHub-hosted runners, where `$HOME` is
job-private and ephemeral. Its **save (post) step prunes the cargo home before
uploading** — both pinned bundles (`400e7407…` v2.7.7, `c1937114…` v2.9.1) walk
`path.join(CARGO_HOME, "registry", "src")` and `rmRF` what they judge unneeded.
On this fleet that is a cross-runner `rm -rf` against a directory another job is
actively compiling from. Cargo launches `rustc` with its cwd inside the
package's `registry/src` dir, so the spawn fails outright, and every sibling unit
still in flight then reports the far more visible `.d` collateral.

**Causal ordering on #974:**

| window | destroyer | victim |
|---|---|---|
| A | `builder-2` rust-cache **save** `02:25:21 → 02:26:32` | `builder-3` 30k property test dies **02:26:02** |
| B | `builder-4` rust-cache **save** `02:50:17 → 02:51:15` | `builder-3` clippy+privacy **02:50:28**, `builder-2` adversarial **02:50:40 / 02:51:15 / 02:51:40** |

Window B kills **two different runners** inside one save. The victims' own
post-steps were `skipped` — they were not destroying themselves.

### A hypothesis that was refuted, not assumed

Cross-runner restore of a `target/` carrying another runner's absolute paths does
**not** explain this: every failing `.d` path names the runner that *owns* the
job (`runner-3` job → `runner-3` path, `runner-5` job → `runner-5` path), and the
fault is in the SHARED `~/.cargo`, not the per-runner `target/`. Such a restore
is genuinely possible — nothing in the cache key distinguishes runner-N from
runner-M — but it is benign: cargo treats a `.d` whose paths are missing as dirty
and rebuilds.

## The fix

All **80** cache steps touching a shared cargo home — 73 `Swatinem/rust-cache`
across 33 workflows, plus 7 `actions/cache` steps whose `path:` covers
`~/.cargo` — now carry:

```yaml
if: runner.environment == 'github-hosted'
```

**Cache-reuse trade-off, measured at zero.** On a *persistent* runner `~/.cargo`
and `target/` already survive between jobs, so the action carries state across a
gap that does not exist here. Worse, its save step also prunes `target/` every
run, so removing it strictly *improves* local reuse. Hosted jobs are
byte-unchanged. Fail-safe: were `runner.environment` ever unavailable it
evaluates false, disabling caching rather than re-enabling the race.

Rejected alternatives: a retry or `continue-on-error` (hides a real corruption),
and a per-runner `CARGO_HOME` in each runner's `.env` (host config, not
repo-gated, 5× registry disk on a Mac with a disk-full history, and a new runner
without it silently re-breaks the fleet).

## Regression lock

New `scripts/validate_shared_home_cache_guard.py`, wired into the `actionlint`
lane. stdlib-only — the fleet cannot provision PyYAML via `setup-python`
(`RUNNER_TOOL_CACHE` is unset). It also FAILS when it finds zero cache steps, so
a broken scanner cannot pass vacuously.

Proven in both directions:
- green on the fixed tree (80/80 checked);
- RED with the guard stripped from the `tenant-path.yml` fuzz step that was the
  measured window-A destroyer;
- RED listing all 80 against the pre-fix tree.

## Mechanism reproduced off-fleet

Throwaway crate, isolated `CARGO_HOME` (the live shared registry was never
touched). Control build passes; the same build with a concurrent
`rm -rf $CARGO_HOME/registry/src` at t+4s emits the identical signature —
`(never executed)` / `os error 2`, then
`could not parse/generate dep info at: …/deps/syn-<hash>.d`, exit 101 — naming
the same crate as the #974 report.

## Gates

- `validate_shared_home_cache_guard.py` — 80/80 OK
- `verify-action-sha-pinning.py` — 545 `uses:` lines SHA-pinned (pins untouched)
- `actionlint` — clean
- `validate_specs.py` — 480 total, all validated
- `validate_secrets_matrix.py` — `code_only=0`
- `check_changelog_duplicates.py` — no duplicated `[Unreleased]` entries

## Not fixed here

`~/.cargo/bin` is shared too, so `cargo install cargo-fuzz` from concurrent fuzz
jobs remains a write race that this guard does not cover.


gmhelmold:
## Follow-up evidence: **restore** is destructive too, and this PR cannot be validated green in isolation

Two things surfaced while watching this PR's own CI. Both strengthen the fix; neither changes it.

### 1. The cache **restore** poisons the shared registry, not just the save

The first analysis pinned the *save* step (`rmRF` over `CARGO_HOME/registry/src`). The restore does it too, by a different mechanism — `tar` extraction unlinks and recreates each file, so a concurrent reader gets `ENOENT` mid-read.

Measured on this PR's run:

| time | event |
|---|---|
| 03:58:29 | `license-policy` job on `corelink-builder-5` (branch `fix/admin-ui-e2e-clerk`) logs `Cache hit` and begins **restoring a 99 MB `~/.cargo` tarball** |
| **03:58:36** | `corelink-builder-3` — clippy + corelink-privacy — dies, `os error 2` |
| **03:59:39** | `corelink-builder-2` — cargo-fuzz smoke — dies, `os error 2` |

Two victims, two different runners, inside one 99 MB extraction over the shared
`~/.cargo`. That step is `Swatinem/rust-cache` with `cache-targets: 'false'`, so
it caches **only** the shared cargo home — the purest form of the bug.

The guard already covers this: `if:` on the step disables **both** restore and
save, so no change is needed. It does mean the mtime-based "was the registry
touched?" check is useless as evidence — `tar` restores original mtimes for both
files and directories.

### 2. Why this PR's four Rust checks can still go red

The destroyer is always **some other branch's job**. `main` and every open PR
still carry the unguarded steps, so while this branch is in review the fleet keeps
being poisoned by jobs that do not have the fix. A fleet-wide fix is not
observable from a single branch — it becomes effective the moment it lands on
`main` and every branch inherits it.

Evidence that the guard itself does what it claims, from this PR's own run:

```
SELF-HOSTED    skipped    Cargo cache (fuzz)
github-hosted  success    Run  Swatinem/rust-cache@…   (×14)
github-hosted  success    Post Swatinem/rust-cache@…   (×11)
```

Skipped on the shared-`$HOME` fleet, unchanged on hosted runners — exactly the
intended split.

Also note the fleet was at **load average 222 / 524 / 575** during this run, which
is why cold rebuilds and the resulting cache traffic hurt so much here.

### Unrelated: `criterion regression gate`

Failed once on `tier_selection_select_starter` (+34% median, `NON_CRITICAL`,
1261 ns → 1690 ns) on a **GitHub-hosted** runner, where this PR is a no-op, and
**passed on re-run**. No Rust or bench commit landed on `main` since the last
green perf run. Nanosecond-scale bench, noisy hosted runner — noise, not a
regression.

PR #1028, opened 2026-08-04
docs(sccache): our published recipe made every cache read remote-only

gmhelmold:
## The defect

Our customer-facing sccache page taught a **remote-pure** WebDAV recipe —
`SCCACHE_WEBDAV_ENDPOINT` + `SCCACHE_WEBDAV_TOKEN` + `RUSTC_WRAPPER`, and nothing
else. Read plainly: **every single cache read a CoreLink sccache customer makes is
an HTTPS round trip, with no local layer at all.** We were not selling "CoreLink in
front of your local cache"; we were selling "CoreLink *instead of* your local
cache", and nobody following the docs had any way to know.

The same block appears in all four locales of the integration page, and again in
the `vs-sccache-s3` comparison page — i.e. it is also what we hand to a prospect who
is evaluating us against their existing sccache + S3 setup.

## The evidence

From sccache's own `storage_from_config` (`sccache/src/cache/cache.rs`):

```rust
if let Some(multilevel) = MultiLevelStorage::from_config(config, pool)? {
    return Ok(Arc::new(multilevel));
}
if let Some(cache_type) = &config.cache {
    return build_single_cache(cache_type, &config.basedirs, pool);   // <- WebDAV lands here
}
// No remote cache configured - use disk cache only
```

sccache selects **exactly one** storage. The disk branch is only reached when *no*
remote backend is configured. So with `SCCACHE_WEBDAV_ENDPOINT` set and
`SCCACHE_MULTILEVEL_CHAIN` unset, the disk branch is unreachable and `SCCACHE_DIR`
is inert — setting it changes nothing.

The fix exists upstream: `SCCACHE_MULTILEVEL_CHAIN="disk,webdav"` builds a two-level
chain (local disk in front of CoreLink, with background backfill of the local layer
on a remote hit). Multi-level chains shipped in sccache **0.15.0**.

## What changed

Every recipe that sets `SCCACHE_WEBDAV_ENDPOINT` now also sets
`SCCACHE_MULTILEVEL_CHAIN="disk,webdav"` and `SCCACHE_DIR`, and states the
sccache >= 0.15 minimum.

- **Prerequisites** now say sccache 0.15+ and tell the reader to check
  `sccache --version` — on an older sccache the chain variable is silently ignored
  and you fall straight back to remote-only.
- **A variable table** says what each of the five variables does. `SCCACHE_DIR` is
  documented as the *path of the local layer*, meaningful only when the chain
  includes `disk` — not deleted, and not presented as if it worked on its own.
- **A `:::caution`** transcribes the `storage_from_config` behaviour above, so the
  reason the chain is mandatory is on the page rather than in a diagnosis doc.
- **"Verify it worked" was wrong under the chain.** The old recipe built twice and
  claimed the warm build proved CoreLink served it — with a local layer present, the
  warm build is served locally and the check proves nothing. It now stops the daemon,
  drops `$SCCACHE_DIR`, and only then does the warm build.
- **The CI example** carries the chain plus `SCCACHE_DIR: ${{ runner.temp }}/sccache`,
  with an honest note that on an ephemeral runner the local layer only helps *within*
  one job.
- **Troubleshooting** gains the two symptoms this defect actually produces: hits are
  counted but every one is a network request, and "`SCCACHE_DIR` appears to be ignored".
- Translated into pt-BR / de / es-419 (no English left behind in the localized files);
  `vs-sccache-s3.mdx` gets the corrected block and a short explanation of why the last
  two variables are not optional.

**No performance number is claimed anywhere.** The local layer is documented as the
correct topology, not as a measured speedup — measuring it is separate work.

## Verification

- `python3 scripts/validate_specs.py` → `469 com schema completo, 11 com YAML only
  (480 total)`, zero failures. (Note: `CLAUDE.md` still quotes `463/0` — that figure
  is stale on `main`; this branch touches nothing under `specs/`.)
- `python3 scripts/validate_docs_reality.py` → `0 drift` (17 pre-existing warnings,
  all unrelated suppression-hygiene entries).
- `vale --config=apps/docs/.vale.ini --minAlertLevel=error docs/integrations/sccache-cargo.md`
  → exit 0. (The Vale gate lints `apps/docs/docs/**` only.)
- Docusaurus build **not** run locally — `apps/docs/node_modules` is absent in this
  worktree and a `pnpm install` there is not cheap on the shared Mac. `docs-ci`
  builds it on `ubuntu-latest`; that job is the proof, not a local claim.

## Out of scope, but found

- `apps/admin-ui/src/components/customer/ConnectClient.tsx` (~line 90) emits the
  **same remote-pure snippet** in the in-product "Connect your client" panel. Not
  touched here — it is a different app and a different gate surface.
- `scripts/e2e-real-client/lib/clients.sh` deliberately forces remote-pure
  (`SCCACHE_DIR=` empty). Left alone on purpose: adding a local layer there would let
  a build be served entirely from disk and never exercise the CoreLink surface the
  probe exists to assert on.


gmhelmold:
Rebasing onto `ed2ab119` — the `okf-wiki-validation` failure on this PR was never caused by it. See #1032: C4b certified main green against an object that the squash-merge of #1022 had already destroyed, a 7-second race between the merge and the push:main checkout.

PR #1030, opened 2026-08-04
fix(admin-ui): the dashboard's sccache snippet used a dead path and no local cache

gmhelmold:
In-product twin of #1028. That PR corrected the **published docs** recipe; this one corrects the recipe the customer copies **from inside the dashboard** — the higher-traffic surface of the two, because it sits on the Connect screen next to their tenant ID.

`apps/admin-ui/src/components/customer/ConnectClient.tsx` emitted two independent defects in six lines.

### 1. The endpoint pointed at a route that does not exist

The snippet built `SCCACHE_WEBDAV_ENDPOINT` as `<origin>/webdav/<tenant>`. There is no `/webdav` route anywhere in the server. The sccache surface is `/cargo/<tenant>/<key>` — `crates/corelink-container/src/routes.rs:118`, mounted via `nest_service("/cargo", …)` at `routes/cargo.rs:264`.

Probed live against prod rather than inferred:

| Path | Status |
|---|---|
| `/webdav/<uuid>/probe` | **404** — no such route |
| `/cargo/<uuid>/probe` | **401** — route exists, demands auth |

Every customer who copied this snippet configured a backend that could not answer a single request. Our own CI (`corelink-reapi.yml`) and the e2e harness (`scripts/e2e-real-client/lib/clients.sh`) already used `/cargo` — the dashboard was the only emitter carrying the phantom path, which is why nothing else caught it. Note this defect is *unique to the dashboard*; the docs were already correct here, so #1028 does not cover it.

### 2. The recipe configured a cache with no local layer

The snippet set the WebDAV endpoint and token and nothing else. sccache's `storage_from_config` only falls through to the disk backend when there is **no** remote backend configured — so with a WebDAV endpoint set, sccache is remote-only: there is no local cache layer at all, every read is an HTTPS round trip, and `SCCACHE_DIR` is inert.

Now emits `SCCACHE_MULTILEVEL_CHAIN="disk,webdav"` plus `SCCACHE_DIR`, and states the sccache **>= 0.15** floor the multi-level chain requires — on an older sccache the variable is silently ignored and you fall back to remote-only. Content and variable ordering are deliberately identical to `apps/docs/docs/integrations/sccache-cargo.md` as corrected in #1028.

### Verification

The emitted snippet was checked by **rendering the component and reading what it produces**, not by reading the source — the template-literal escaping is the exact thing that would silently break here:

```
export SCCACHE_WEBDAV_ENDPOINT=https://corelink-api.humangr.com/cargo/<tenant>
export SCCACHE_WEBDAV_TOKEN="${CORELINK_TOKEN}"
export SCCACHE_MULTILEVEL_CHAIN="disk,webdav"   # local disk in front of CoreLink
export SCCACHE_DIR="$HOME/.cache/sccache"       # where the local layer lives
export RUSTC_WRAPPER=sccache
```

`$HOME` stays literal, and the token stays a `${CORELINK_TOKEN}` reference — never inlined (CTRL-CRED-001). The probe test was scratch and is not committed.

Gates run locally in a clean worktree:

- `pnpm lint` — **0 errors** (10 pre-existing warnings, all unused `eslint-disable` directives in files this PR does not touch)
- `pnpm typecheck` — clean
- `pnpm test` — **607 passed / 0 failed** across 115 files; no snapshot covers this snippet, so none needed updating
- `python3 scripts/validate_specs.py` — 480 total, all valid
- `python3 scripts/validate_docs_reality.py` — 0 drift
- `python3 scripts/check_changelog_duplicates.py` — OK

No restyling, no refactor: the panel, the `origin`-derived endpoint mechanism, and every other surface's snippet are untouched. No `localStorage.setItem`, no `console.log`, no absolute client paths added.


gmhelmold:
Rebasing onto `ed2ab119` — the `okf-wiki-validation` failure on this PR was never caused by it. See #1032: C4b certified main green against an object that the squash-merge of #1022 had already destroyed, a 7-second race between the merge and the push:main checkout.

PR #1042, opened 2026-08-04
fix(worker): finish #1036's review — the test harness's no-delay case, and a fail-closed control the repo documented as fail-open

gmhelmold:
#1036 was squash-merged while its third review round was still open, so three findings never landed. This PR is those, plus everything five further review rounds turned up on top of them.

## The test harness (from #1036's round 3)

**MED-E — the "no delay" case had a 150 ms delay.** The no-delay sentinel was a *member* of `SlowTarget` (`"none"`), and `phaseOf` returned that same `"none"` for any statement it could not classify. So `makeEnv("none")` delayed **every unclassified statement**, including `SELECT tier FROM tier_selections` inside the `wdb` window. The accounting assertion added for that case says *"`wdb` and the sum are both near zero"* — it measured `wdb ≈ 151 ms`. After the fix: 11 ms. The sentinel now lives outside the union (`SlowTarget | null`).

**MED-F — an assertion that fixing MED-E would have made vacuous.** `qtier < 50` in the cache-warm case now runs against `makeEnv("tier")`, so a cache *miss* costs a declared 150 ms. Proven RED by disabling the L1 tier cache: `expected 150 to be less than 50`.

## A fail-closed control the repo documented as fail-open

`resolveConsumerKey` **refuses** the shared-key fallback for a dedicated key that is SET but under the length floor — it logs *"REFUSING to fall back … that would silently widen the blast radius"*. Its own doc-comment claimed the opposite: *"A too-short dedicated key is treated as ABSENT (falls through to the shared key)"*, and that phrasing had been copied into the module header, the `Env` declaration and the live `/_internal/*` gate comment.

Concrete risk: someone simplifying the function to match its documented contract reintroduces exactly the widening the code was written to prevent. **All five corrected.** Both `worker/src` commits here are provably comment-only.

## What the shared key actually is

The OKF `worker-edge` concept now states it, enumerated exhaustively **and shipping the grep that re-derives the enumeration** — because a count in prose is a claim, and five earlier drafts of this paragraph were each wrong in a different direction (overstating the blast radius, then understating fail-closed coverage, then undercounting the mechanisms twice).

`CORELINK_INTERNAL_AUTH_KEY` is read in three **roles**:

1. **Inbound gate credential**, both fail-closed — `resolveConsumerKey` (four gates, per-consumer, refuses set-but-short) and `requireInternalAuth` (two gates, shared-key only, **no per-consumer isolation at all**, so a shared-key holder passes outright — the widest inbound exposure). Six inbound gates total.
2. **Outbound credential presented onward** — four `?? ` sites, plus the `onboarding`/tier-select-checkout arm which reads the shared key **directly with no dedicated-key preference**, so provisioning `CORELINK_PAT_MINT_AUTH_KEY` can never narrow it. On the money path.
3. **The multi-region fan-out marker itself** — which is what the note this all hangs off is about.

## Process notes worth keeping

- Resolving the #1043 conflict by taking main's concepts **wholesale** was right for citations but silently reverted this PR's own checkpoint-repoint commit, making it a no-op inside its own PR. Frontmatter has to be re-derived after a wholesale take, not assumed to survive.
- My re-anchor helper matched `worker/src/index.ts` as a substring of `apps/signup-worker/src/index.ts` and corrupted three unrelated concepts. Reverted, anchored with `(?<![\w./-])`, and the three are byte-identical to main.
- Two v6 citations were written from a grep taken before the edit that shifted the file, and landed on plausible-looking but wrong code. Caught by reading the cited lines, not by a gate.
- **`validate_okf` was green through every one of these.** C5 compares a file against its own blob anchor, so once the anchor advances it compares the file to itself; C4 only *warns* on an orphaned checkpoint. Green repeatedly meant "not consulted", not "correct".

## Gates

worker **629/629** · `validate_okf` 160 concepts, 0 stale, 0 drift · `validate_specs` 469+11/480 · `shellcheck` clean · all six checkpoints verified reachable · `grep unset/short` empty


gmhelmold:
Superseded by #1048.

#1047 replaced `qmeter`/`qstor` with a single batched `qbatch` phase. This branch's test asserts the retired labels, its `index.ts` conflicts with the batching, and its 7d paragraph documents an oracle #1047 deliberately retired — so it cannot be merged or rebased, only re-derived.

All three still-open defects are carried forward to #1048 against the `qbatch` base, and they matter: #1047 was authored from a base predating this PR, so it **re-introduced the `"none"` sentinel collision verbatim**. That bug, the assertion it made vacuous, and the fail-closed-control-documented-as-fail-open are live on `main` right now.

Nothing is lost. Closing rather than force-merging because a "take ours" resolution would have reverted #1047's batching, and a "take theirs" would have left the tests red.

PR #1048, opened 2026-08-04
fix(test): the wdb sub-phase harness's "no delay" case had a 150ms delay — and the suite passes either way

gmhelmold:
Supersedes #1042, which can no longer be merged: #1047 replaced `qmeter`/`qstor` with `qbatch`, and that branch's test asserts phases the Worker no longer emits. Same defects, re-derived onto the `qbatch` base.

#1047 was authored from a base predating #1042, so it **re-introduced defect 1 verbatim** and left 2 and 3 untouched. All three are live on `main` right now.

## 1. The harness's "no delay" case has a delay in it

The no-delay sentinel was a **member** of `SlowTarget` (`"none"`), while `phaseOf` returned that same value for any statement it could not classify. So `makeEnv("none")` delayed **every unnamed statement** — including `SELECT tier FROM tier_selections` inside the `wdb` window, and the PAT lookup.

Measured on this base, the supposedly-quiet case:

| | split |
|---|---|
| colliding sentinel | `{"auth":152,"wdb":154,"qtier":152,…}` |
| sentinel outside the union | `{"auth":1,"wdb":2,"qtier":1,…}` |

A case that claims to inject nothing was injecting 152 ms into two phases.

**The suite passes both ways.** This is a correctness bug in the *instrument*, invisible to the instrument's own assertions — which is exactly why it survived two review rounds and a re-introduction, and why it had to be **measured** rather than asserted. Fixed by moving the sentinel out of the union (`SlowTarget | null`), so "delay nothing" and "cannot classify this" can never again be the same value.

## 2. An assertion that fixing 1 would have made vacuous

With 1 fixed, the cache-warm `qtier < 50` holds regardless — an un-delayed D1 read also costs ~0 here. It now runs against a declared tier delay. Proven RED by disabling the L1 tier cache: `expected 151 to be less than 50`.

## 3. A fail-closed control documented as fail-open

`resolveConsumerKey` **refuses** the shared-key fallback for a dedicated key that is SET but under the length floor, logging *"REFUSING to fall back … that would silently widen the blast radius"*. Its own normative contract claimed the opposite — *"A too-short dedicated key is treated as ABSENT (falls through to the shared key)"* — and that phrasing was copied into the `Env` declaration and the live `/_internal/*` gate comment.

Concrete risk: someone simplifying the function to match its documented contract reintroduces exactly the widening the code exists to prevent. No behaviour change.

## What the shared key actually is

Now stated in the `internal_auth.ts` module header — with the code, not in a phase concept whose hook #1047 retired — enumerated **by role** and **shipping the grep that re-derives the enumeration**, because a count in prose is a claim and this one was wrong in five successive drafts:

- **Inbound gate credential**, both arms fail-closed: `resolveConsumerKey` (four gates, per-consumer) and `requireInternalAuth` (two gates, **shared-key only, no per-consumer isolation at all** — a shared-key holder passes outright, the widest inbound exposure). Six inbound gates.
- **Outbound credential presented onward**: four `??` sites, plus the `onboarding`/tier-select-checkout arm which reads the shared key **directly with no dedicated-key preference** — so provisioning `CORELINK_PAT_MINT_AUTH_KEY` can never narrow it. On the money path.
- **The multi-region fan-out marker.**

## Gates

worker **646/646** · `validate_okf` 160 concepts, 0 stale, 0 drift · `validate_specs` 469+11/480 · tsc unchanged at 14 pre-existing

Opened as a **draft**: an automated sweep merged #1036 mid-review earlier today. Undrafting once the independent review returns.


gmhelmold:
Round-2 review applied — N1 through N6, every edit asserted on its target first (a silent no-op `str.replace` is what produced the two "reported fixed, absent from the diff" findings).

Three follow-ups deliberately **not** folded in, so they don't ride along unreviewed:

1. **`credential-handling` has no `source_blobs`.** Its C5 compares cited-line content between the checkpoint commit and the worktree, and the checkpoint is the commit that finalized the file it cites — so the comparison is byte-identical by construction and the gate can say nothing. It is also the only concept in this set with no C4b backstop after a squash. Durable fix: give it `source_blobs`.
2. **No CI lane runs `tsc` on `worker/`.** `worker-vitest.yml` runs only vitest, and `worker/tsconfig.json` excludes `tests/` — which is why the type half of this PR's own headline fix shipped broken and nothing went red.
3. **403 vs 503 is inconsistent in the code itself** (pre-existing): the `index.ts` edge gate returns 403 for an unbound key while the `internal_auth.ts` gates return 503, and `runner_mint.ts:214` documents 403 where the code returns 503.

Also noted and left alone: `docs/okf-wiki-site/index.html` is a stale generated rendering (regen via `scripts/okf_render.py`), and the sealed dated audits carry pre-relocation ranges by design — point-in-time artifacts.

gmhelmold:
## Follow-ups from the merge round — none blocking this PR

### 1. `tsc` on `worker/` does not complete — the explosion is in `tests/`, not `src/`

`tsc -p worker/tsconfig.test.json --noEmit` ran **1h46m at 98% CPU (89 min of CPU time, RSS 702 MB)** without finishing, on a machine that is also the self-hosted CI fleet. Killed it.

It is not a config problem: `skipLibCheck` is on, and of the 733 files pulled in, 677 are `node_modules` type deps of only **56 source files**. `--listFilesOnly` returns instantly, so resolution is fine — it is the checker.

Bisected: **`src/` alone compiles in 10.8 seconds.** Adding `tests/` is what blows it up. Three test files landed on main this cycle (`server_timing_origin_subphases.test.ts`, `do_d1_probe_route.test.ts`, `durable_object.test.ts` +282), which is where I would look first.

### 2. `src/` has 14 type errors — all of them already on `main`

| | |
|---|---|
| `TS4111` | 12 |
| `TS2379` | 1 |
| `TS2322` | 1 |

In `src/replication_coordinator_do.ts` and `src/index.ts`. I ran the same check in a detached worktree at `origin/main`: **identical set, 14/14, zero introduced by this PR.**

Both of these survive because **no CI lane runs `tsc` on `worker/`** — the follow-up filed earlier on this PR. It is no longer hypothetical: it is why 14 type errors sit on `main` unnoticed and why a test suite that cannot be typechecked in under two hours went unremarked.

### 3. Retraction

I earlier reported `tsc` green at `6e398b8b`. That was wrong. The job produced an **empty output file** with exit 0 — not even the `echo TSC_EXIT` I had appended — because the scratchpad worktree was cleaned out from under it. A second job I killed also reported exit 0. Aborted job, success exit code; I accepted the exit code without reading the output. There is no valid `tsc` result for `tsconfig.test.json` on this branch, and given #1 there cannot be one until the `tests/` explosion is fixed.

gmhelmold:
## Merging with one red gate — documented reason, owner-authorized

`trivy fs` is RED with 5 HIGH CVEs. Merging anyway, with the owner's explicit approval, because **this PR provably cannot affect them**:

- `git diff --name-only <merge-base> HEAD` touches **no** `package.json`, no `pnpm-lock.yaml`, no `package-lock.json`. Same for the merge commits. The dependency tree here is byte-identical to `main`'s.
- `trivy fs` **passed on this same PR** before the merge and fails after. The delta is time, not content: these CVEs were published against dependencies that were already on `main`.
- `trivy` runs on pull_request only, never on push to `main`, so `main` shows no red — but any PR opened right now fails identically.

| dep | reached via | fixable |
|---|---|---|
| `image-size@2.0.2` — CVE-2025-71329 / -71330 (DoS via crafted image/ICNS) | `@docusaurus/mdx-loader` (docs-site build) | **No** — 2.0.2 is the latest published version |
| `nanoid@3.3.12` — CVE-2026-67213 / -67214 (infinite loop) | `postcss` (build) | Yes — 3.3.17 |

Neither is reachable from the Worker, the container, or any runtime path; both are build-time only.

This is **not** a waiver of the gate and not a claim the CVEs are acceptable. It is a scope call: the finding is repo-wide and pre-existing, and holding a correctness fix hostage to an unrelated supply-chain issue with no upstream patch helps nobody. A dedicated PR follows with the `nanoid` override and a decision on `image-size`, which needs an owner call because there is nothing to upgrade to.

Other 10 checks green: vitest, dco, gitleaks, okf-wiki-validation, secrets-matrix drift, CHANGELOG, claims-vs-reality, labels, size, dependabot sentinel.

PR #1060, opened 2026-08-08
fix(container): call Cloudflare's own idle auto-destroy (setInactivityTimeout)

gmhelmold:
## The one-line version

Cloudflare has a native idle reaper. We documented it and never called it.

`Container.setInactivityTimeout(ms)` (`worker-configuration.d.ts`, `interface Container`) makes **workerd** destroy an idle container, with no help from this Worker. It is named in `worker/src/durable_object.ts`'s header doc-comment at line 18 — and `grep -n setInactivityTimeout` on that file returned exactly one hit: **the comment**.

Instead the repo hand-rolled the alarm reaper in `alarm()`. It shipped in #927 on 2026-08-02 and **has never been observed to work**.

## Evidence

Measured 2026-08-08 against the **per-application** CF Containers API — the list endpoint serves stale reads and misled me twice today, so per-app is the only trustworthy source:

| region | instances | health |
|---|---|---|
| `syd` | 7 | `active: 0, healthy: 7` |
| `nrt` | 7 | `active: 0, healthy: 7` |

Seven containers up in regions with **zero traffic**, 35 fleet-wide (5 × 7 × 4 GiB = 140 GiB resident) — the same 35 the #927 note recorded three days after that fix deployed.

Idle **memory**, not vCPU, is what bills. Invoice IN-72804238 (Jun 25 – Jul 24): Container Memory **$84.09** vs vCPU **$5.28**, against **$0.00** of Workers / R2 / D1 traffic.

## Why platform-side, given we already have a reaper

It survives DO eviction, a broken alarm chain, and a wedged isolate — precisely the failure modes that made containers immortal. The #927 note's own leading hypothesis was that the immortal DOs never get evicted, so they never receive the fixed code. A platform timer does not care.

**The alarm reaper is kept.** Defence in depth, and it also stops re-arming the chain so the DO itself can hibernate (a live alarm chain bills DO duration on its own — both halves matter).

## The unspecified bit, handled

The type declaration has **no doc-comment**, so it is genuinely unknown whether the timer resets on activity or is an absolute deadline from arming. Under the absolute reading, arming once at start would kill a **busy** container mid-request one window later.

So it is armed at start **and** re-armed from `alarm()` while the container is non-idle — correct under both readings. Re-arming rides the existing 30 s alarm and **never** the request path; a per-request `await` would tax the hot path #1055 just shrank.

Arming failures are logged, never thrown. A container that cannot arm its idle timer must still serve — but it must not fail *silently*, because a silently unarmed timer is exactly how this leak survived a whole fix cycle.

## Tests

`worker/tests/durable_object.test.ts` — 62 pass.

The load-bearing one is negative-controlled: neutralising **both** call sites makes `re-arms with IDLE_TIMEOUT_MS on an alarm tick for a live, non-idle container` **fail** (verified, then restored). The other two assert degradation — a runtime without the API, and an arming throw — and pass either way by design.

No new `tsc` errors; the 14 reported are the pre-existing `main` baseline in `replication_coordinator_do.ts`.

## What this PR does NOT claim

**It does not prove the bill is fixed.** #927 looked correct and did nothing. The only acceptable proof is the live instance count falling after deploy — and prod containers are currently at `max_instances: 0` deliberately, to stop the burn while this lands. Measuring that is my job, not this PR's.


gmhelmold:
Admin-merging with one red check, documented:

**`trivy fs` — pre-existing, not introduced by this PR.**
- Trivy runs **only on `pull_request`** (no main lane), so it fails on all three of today's PRs (#1060, #1061, #1062) with entirely unrelated diffs — the signature of a pre-existing dependency-tree finding, not a regression.
- This PR's diff touches only `worker/src/durable_object.ts`, its test, the OKF concept and CHANGELOG — **no `package.json`, no lockfile.** It cannot have introduced an npm CVE.
- The findings are 5 HIGH in `image-size` and `nanoid` (transitive). `nanoid` has a fix; **`image-size` CVE-2025-71329/71330 has no fixed version published** (empty 'fixed' column), so the gate cannot go green today regardless of this PR.

Everything load-bearing is green: 62 vitest (idle-reaper re-arm negative-controlled), okf-wiki 160/0/0 after a surgical reconcile, dco, changelog, secrets-matrix, gitleaks.

The trivy finding is tracked separately (pnpm.overrides bump for image-size once a fix ships).

PR #1061, opened 2026-08-08
chore(ci): move 8 cron-only workflows off hosted runners

gmhelmold:
Owner rule: **zero GitHub-Actions hosted spend** — self-hosted mac or CoreLink runners only.

## Where the money actually goes

Measured 2026-08-08, one day of runs:

| trigger | runs |
|---|---|
| `schedule` | **50** |
| `pull_request` (+ `_target`) | 13 |

**Cron dominates the hosted bill, not PRs.** Single biggest spender: `cas-canary` at `17 * * * *` — hourly, 24 runs/day, hosted.

## Selection is deliberate, not find-and-replace

Each of these 8 is **cron-only** (no `pull_request`/`push` trigger, so no PR gate moves onto the founder's Mac) **and** free of Linux-specific steps — the only `ubuntu` match in each file was the `runs-on` line itself, verified per file rather than assumed.

`ac-bucket-acl-cron` · `billing-reconcile-daily` · `gc-sweep-dry-run` · `okf_nightly` · `semgrep` · `compliance-weekly` · `dr-drill-monthly` · `byok_kill_switch_drill_weekly`

Four of those are **daily**, so this is recurring spend, not one-off.

## Deliberately NOT moved

- **`cas-canary`** — `ubuntu-latest` on purpose: it needs a **datacenter IP** to imitate a customer's CI. Converting it breaks the thing it exists to test. Ironically the biggest spender; it needs a different fix (frequency, or CoreLink runners), not a `runs-on` swap.
- **`codeql`, `coverage`** — pass both safety rules but are heavy CPU. Nightly CodeQL and weekly coverage on the founder's Mac trades a money problem for a machine-stability problem; that Mac has crashed under CI load before.
- **`ffi-matrix`, `cas_foundation`, `backup-daily`, `byok_matrix`, `reproducible-build`, `mutation-nightly`, `s10-ship-gate`, `audit-chain-daily-verify`** — real `apt-get`/`docker` steps. A `runs-on` swap would break them silently.
- **The 16 that also trigger on `pull_request`/`push`** — moving those relocates PR-blocking gates onto the Mac. Separate decision, separate PR.

## Verification

YAML re-parsed after the edit; every job resolves to `[self-hosted, mac, corelink-builder]`, matching the format already used by the working `e2e-clerk-signup` / `e2e-stripe-checkout` / `e2e-admin-ui-render`.

⚠️ **No CI ran on this PR.** GitHub Actions is billing-blocked — jobs fail in ~2 s with `steps=0`, which is the "job never started" signature, not a test failure. Review this by reading the diff, not by waiting for a green check.


gmhelmold:
Superseded by #1062, which moves **all 104** hosted jobs (not just these 8) onto `corelink`.

`corelink` is the better target than the mac fleet for these files: it is Linux, so nothing depends on the per-file "no apt-get/docker" screening this PR needed. #1062 also takes the ones this PR deliberately skipped — `cas-canary` (needs a datacenter IP; CoreLink runners are on CF Containers, so they qualify), `codeql` and `coverage` (too heavy for the founder's Mac, fine on a 4 vCPU box), and the 16 that also trigger on `pull_request`.

Closing unmerged. Nothing here is lost — it is a strict subset of #1062.

PR #1062, opened 2026-08-08
chore(ci): move every hosted job to CoreLink runners — zero Actions spend

gmhelmold:
Owner rule, stated three times on 2026-08-08: **no GitHub-Actions hosted billing.** Everything on self-hosted or CoreLink runners.

## Result

| runner | jobs |
|---|---|
| `corelink` | 132 |
| `[self-hosted, mac, corelink-builder]` | 80 |
| `[self-hosted, Linux, X64]` | 10 |
| **GitHub-hosted** | **0** |

104 jobs across 61 files moved off `ubuntu-latest` / `ubuntu-x64-4core`.

## Why `corelink` and not the mac fleet

**It is Linux.** The workflows with `apt-get` / `docker` steps keep working. Sending them to a macOS runner would have broken them *silently* — worse than costing money, because a broken scan looks the same as a passing one until you read it.

It also adds no load to the founder's Mac, which has crashed under CI before.

The three `ubuntu-x64-4core` jobs came along: they were on a bigger hosted box because the 2-core default OOM'd on link (`ld Bus error, signal 7`). The CoreLink box is 4 vCPU / 12.5 GB, which clears that bar.

Trailing comments on each `runs-on` are preserved — several record exactly why that job needed a particular box.

## One judgement call, not a regex

`secrets-drift.yml` deliberately kept its **schedule** arm hosted, arguing the daily run is SOC 2 CC6.1 evidence and a silently-missed run is an invisible evidence gap.

Now `corelink`. Two things make that acceptable: SOC 2 is **parked**, so this is not live audit evidence; and the feared failure mode is *silence*, which is a **monitoring** gap, not a runner gap — equally invisible on hosted when the workflow is disabled or billing is blocked, **both of which happened this week**. If that evidence stream becomes load-bearing again, the fix is a "no successful run in 24 h" alert, not a paid runner.

## Deliberately left alone

`cargo-cache-latency-probe.yml` can still put `ubuntu-latest` in its matrix — but only when a human passes `include_hosted=true`. It exists to benchmark the CoreLink box *against* hosted, never fires on cron, and costs nothing unless explicitly requested.

## Two risks worth knowing now

Neither costs money; both cost debugging time.

- **Concurrency.** The fabric's real ceiling is **~7**, not the nominal 20. A burst **drops** a job — it surfaces as a *missing* run, not a red one.
- **Image drift.** The box image has changed underneath us before (rustc 1.91 → 1.96 in two days). A version-sensitive build can break with no commit.

## Verification

All 222 workflow YAMLs re-parse clean after the edit.

⚠️ **No CI ran.** Actions is billing-blocked — jobs fail in ~2 s with `steps=0`, the "never started" signature, not a test failure. Review by reading the diff.


gmhelmold:
Closing unmerged. The approach is sound but the target image is not ready.

**Blocker: the `corelink` runner image is not a drop-in for `ubuntu-latest`.** Proven on this PR's own CI run once billing was restored:
- **no `python3`** — `secrets-matrix`, `drift check`, `breaking-change detection` and most gates in this repo are stdlib python3 scripts; they died with exit 127 (`python3: command not found`).
- **no Docker** — `runner-probe.yml` run 30865752338 printed `docker ABSENT` (already handled here by keeping the 3 image-building jobs hosted).
- **wrong rustc** — the image ships 1.96.0; the workspace pins 1.91.1 (ADR-0015). Handled here by keeping the 35 Rust-provisioning jobs hosted, but it confirms the image drifts from what CI needs.

So 'zero hosted Actions' is an **image** problem, not a workflow problem. The fix lives in the `corelink-runners` repo: bake python3 + Docker + the pinned toolchain into the runner image, re-run `runner-probe.yml` to prove they're present, THEN redo this migration against a known-good image.

Not lost — the mechanical migration (94 jobs → `corelink`, with the Docker/Rust carve-outs documented at each `runs-on`) is easy to reproduce once the image is ready. Nothing here shipped, so main is unaffected.

PR #1173, opened 2026-08-19
chore(container): raise prod cache max_instances 10-20 → 2000 (account ceiling)

gmhelmold:
## What

Removes the arbitrary low per-app container cap on the 5 prod cache workers (`prod` + `prod-{sam,lhr,nrt,syd}`): `max_instances` 10-20 → **2000** (the account's DO-container ceiling).

## Why

- The cache container is **1-per-active-tenant** and **idle scales to zero**, so a high ceiling costs **nothing when idle** — it only bounds worst-case concurrent-active, which is paying tenants using the product.
- An app cap **below** the account limit just **starves** real concurrent tenants for no saving (old 10/region trapped past ~7 concurrent active tenants).
- The real backstop is the **account limit** (`durable_object_offset_instances = 2000`, CF-enforced, shared across all container apps). Setting each app to that ceiling means the account limit is the only bound — not an invented number.

## vCPU note

The account vCPU limits (`total_vcpu = 1500`, `vcpu_per_deployment = 4`) are **not** binding here: 2000 `basic` instances ≈ 125 vCPU, far under 1500. Raising the account limits (more total vCPU, or a bigger per-instance type) is a **Cloudflare support request**, tracked separately.

## Scope / safety

Staging (10) + dev default (5) unchanged. No behavior change at current load — pure headroom. `wrangler --dry-run` clean. Deploy: `cf-deploy-prod` (all 5, `skip_pin_freshness=true`, container recycles to the same live `77c371f8-r1`).


gmhelmold:
Heads-up: this `max_instances → 2000` bump is **rejected by the live Cloudflare API** (it passes `--dry-run` only, which never calls the quota endpoint). `cf-deploy-prod` run 32225170810: `prod-nrt`+`prod-sam` applied, but `prod`/`prod-lhr`/`prod-syd` failed on:

> `Surpassed total account limits: ... instances of count 2000`

That split the fleet across image versions and blocked the F3.2 image roll, so I restored the deployable baseline (`prod=20`, regionals=`10`) in #1176 and re-rolled all 5 regions onto it. The headroom idea is right — it just needs the **Cloudflare support request to raise the account vcpu/instance quota first** (as this PR's body already flagged "tracked separately"); once that quota exists, a deployable number can re-land. No action needed from you unless you want to prioritize that CF quota ticket.

PR #1250, opened 2026-08-24
docs(backlog): the archive found a chain fork on 2026-08-14

gmhelmold:
## B-025 — eight audit-chain partitions forked on 2026-08-14

The archive built for B-015 refused eight partitions, and the reason is an integrity defect it **inherited, not caused**. `_public`/`wnam` sat at exactly **131** unarchived rows for eight consecutive hourly ticks while `enam` drained 29,832 → 13,097. Pulling that partition out of D1 and running the verifier over it:

```
AUDIT_CHAIN_BREAK_DETECTED: tenant=_public: region=wnam:
  sealed archive chunk sequence gap: expected 11, found 10
```

Six sequence numbers (10, 13, 20, 30, 32, 37) are each held by **two** rows, and the six numbers immediately after them (12, 14, 21, 31, 33, 38) are missing. Each pair is the `read.attempted`/`read.served` pair of one request, and the two rows carry **different `prev_hash`** values — the chain forked, it did not merely duplicate a label.

**Measured scope:** 8 of 360 partitions, 1,505 excess rows, 3,010 rows on a duplicated sequence.

**Historical and bounded.** Every one of those 3,010 rows was sealed inside a single 17-minute window, 2026-08-14 01:43:40 → 02:00:50 UTC. The drain on `main` today aborts a partition when the head drifts under it rather than forking. Nothing here suggests tampering — a tamperer does not leave both branches behind.

**Why nothing noticed for ten days.** The daily verifier only examines chunks that reached the bucket, and these never did. The B-021 absence monitor stays quiet as long as the healthy partitions keep writing. That is **B-022, now with a live instance instead of a hypothesis**.

Remediation stays **open for the owner** — re-sequencing sealed rows rewrites the evidence the chain exists to protect, and the options differ in what they claim to an auditor.

## B-017 → done

`bazel-starter-ci` is `active` again and ran green today at 10:48–10:52 UTC. Green means something here: the verify step's own zero-entries branch is `sys.exit(1)`, so a passing job cannot mean the cache was unreachable. It printed:

```
Cache hit ratio: 3/3 = 100.0%
```

Two caveats recorded rather than smoothed over: three actions is a small sample for a surface this close to the product claim, and the same run measured `COLD_MS: 8954` against `WARM_MS: 11450` — the warm build was **slower**. Nobody should quote this example as a speed proof.

The bazel waiver is removed; it now named an active workflow, which the guard correctly treats as drift in the other direction. Guard re-run: **PASS**, two remaining waivers both unexpired.

`python3 scripts/backlog_verify.py` — 25 items, confirmed=25, drifted=0, stale=0, broken=0.


gmhelmold:
**Correction pushed.** I closed B-017 on a run whose ref I had not checked.

Every green `bazel-starter-ci` run today — including the `Cache hit ratio: 3/3 = 100.0%` quoted in the description above — is on the branch `fix/bazel-starter-cache-hit-parser`, i.e. **PR #1249, still open**. That PR exists precisely because `main`'s verify step counts a field Bazel stopped emitting, so on `main` the parser still sees zero entries.

So: the credential half (#1238) is genuinely closed and the workflow is `active` again. The cache half is evidence about a **branch**, not about the gate this repo ships. B-017 goes back to `open`, and its `verify` now greps for the stale `remoteCacheHit` field so it cannot be closed until the parser actually changes on `main`.

The B-025 chain-fork finding is unaffected — that was measured directly against prod D1.

gmhelmold:
Rebased onto main after #1249 landed.

**B-017 dropped from this PR.** Main's own closure is better and more accurate than mine — it names the real root cause (the job parsed `remoteCacheHit` while the execution log carries `cacheHit`), and it caught something I did not: `.bazelrc` registered an **unscoped** `--credential_helper`, so every build was sending the tenant PAT to `bcr.bazel.build`. Its numbers also supersede the warm-slower reading I quoted, which came from a run before that fix. My commit reopening B-017 is skipped.

**What remains here is B-026** (renumbered from B-025, which main took meanwhile): the eight forked audit-chain partitions, measured directly against prod D1 and unaffected by any of the above.

Also unwedged the Mac CI in passing — all four `corelink-builder` runners reported `busy=true` with zero jobs in progress and seven runs queued behind them, one since 05:04. A launchd restart plus cancelling the orphaned queue cleared it; a dispatched `rustfmt` then ran green, which is the proof that assignment works again. The `busy=true` flag itself is still stale on GitHub's side and appears to be cosmetic.

PR #1283, opened 2026-08-24
perf(worker): attribute the 150 ms hole in `wdb` — gated `qdo` + `qother`

gmhelmold:
## The hole

`wdb` (`index.ts:3965`) spans `stOriginStart - stAuthEnd` — everything between auth and origin — but publishes only three sub-phases. Measured in production today (`GET /v1/ac/{tenant}?limit=1`, tenant `d863fafb…`, **free** tier, colo GRU, n=8):

| | `wdb` | Σ(`qtier`+`qbatch`+`qresid`) | unattributed |
|---|---:|---:|---:|
| n=8 | 268-287 ms | 115-130 ms | **144-157 ms** |

On 2026-08-04 that residue was **0 on 30 of 30** requests (`lib/quota.ts:657-661`). It appeared since, and nothing on the wire says where it went.

The suspect is in the same window and untimed: `await serveViaDO(...)` (`index.ts:3267`), live because `EDGE_DO_METER = "serve"` is set on prod and all five regions (`wrangler.toml:310,746,896,1055,1208`).

A controlled A/B in prod supports it. Same endpoint, same colo, same minutes — only the tier differs:

| | free (`d863fafb…`) | enterprise (`…f0006`) |
|---|---:|---:|
| `wdb` | 268-287 ms | **0-8 ms** |
| `qbatch` | 115-124 ms | **absent** |
| unattributed | 144-157 ms | **zero** |
| total | 579-721 ms | 254-294 ms |

An uncapped tier fails `serveGateActive` (`lib/edge_do_meter.ts:212` — `cap !== MAX_SAFE_INTEGER`) so the DO hop never runs, and `storageCapIsFinite` is false so `runQuotaBatch` issues no round trip at all (`ranD1:false`). Both vanish together, and the residue vanishes with them.

This PR does not fix that. It makes it readable.

## What it does

- Times the `serveViaDO` hop into `stQDoMs`, stamped in a `finally` so a fail-open DO outage still accounts for the wall time it cost.
- Publishes `qdo` and `qother` (the `wdb` residue), reusing `ohop`'s reconciliation discipline verbatim: `-1` phases excluded from the sum, never counted as 0; a sum overshooting `wdb` is dropped whole as `qother;dur=<wdb>;desc="unreconciled"` rather than silently redistributed.
- Invariant when on: `qtier + qdo + qbatch + qresid + qother === wdb`, exactly, always.

## Why it is behind a flag

`qdo`'s presence/absence is a confirmation oracle for the internal auth key.

`serveGateActive` requires `meter === true`; `meter = !isFanout && requestQuotaEnabled` (`:3225`); and `isFanout` is a constant-time compare of the **client-supplied** `x-corelink-fanout-from` header against `CORELINK_INTERNAL_AUTH_KEY` (`:3189-3194`). So on a capped tier, a correct key guess yields `qdo` absent while a wrong one yields it present — one bit per request, and precisely the oracle that the `qmeter`+`qstor` → `qbatch` merge exists to close (documented at `:3409-3416`).

So both new phases are gated on `SERVER_TIMING_WDB_DETAIL === "on"`, off by default and deliberately **not** in `wrangler.toml`. With the flag unset the header is byte-identical to `main` — asserted by a test. The clock is stamped unconditionally; only the emission is gated.

## Verification
`vitest 820/820`. `tsc --noEmit` errors 14 → 14 (the 14 are pre-existing on `main`). No behaviour change: no reordered awaits, no new awaits, no changed responses.


gmhelmold:
**OKF reconciliation pushed** (`51bc66a`) — the gate was red because this branch shifted line numbers in `worker/src/index.ts`, leaving six concepts citing the wrong lines. Baseline check first: `origin/main` reconciles to **0 stale**, so every stale concept here was caused by this branch and the scope was closed.

97 citations repaired across `auth/pat-moat`, `launch/money-path`, `planes/durable-object`, `planes/request-flow`, `planes/worker-edge`, `tenancy/isolation`. Two concepts gained real content rather than just coordinates: `planes/request-flow` documents `qdo`/`qother` inside the `wdb` window, and `planes/worker-edge` documents why they are gated.

`validate_okf.py` green, `okf_reconcile.py` 0 stale, `okf_status.py` green.

---

**One pre-existing issue found while verifying, deliberately NOT fixed here.**

The frozen contract says to cite the implementing line, never a doc-comment. Measuring the concepts touched by this PR:

| | citations to `worker/src/index.ts` | landing on a comment line |
|---|---:|---:|
| `origin/main` | 83 | **18** |
| this branch | 88 | **18** |

Same absolute count — the reconciliation faithfully remapped what was already there, including citations that already pointed at comments. So this is pre-existing wiki debt, not a regression from this PR, and `validate_okf.py` does not check it (it is a judgement rule, not an automated one).

Repairing 18 unrelated citations inside an instrumentation PR would be scope creep, so they are left alone and reported here instead.

gmhelmold:
**Correction to my earlier comment — the "pre-existing wiki debt" I reported does not exist.**

I claimed 18 of 88 citations to `worker/src/index.ts` land on comment lines rather than implementing lines, in both `main` and this branch, and called it pre-existing contract debt.

That was a measurement error. My check extracted the leading number from every citation without distinguishing a **point** citation (`file.ts:1234`) from a **range** (`file.ts:1234-1290`). A range that opens on a doc-comment and spans the implementation below it is not the defect the contract describes — the contract forbids pointing AT a comment *instead of* the code, which a range covering the code does not do.

Re-measured on merged `main`, over the same six concepts, separating the two forms:

| | count | landing on a comment |
|---|---:|---:|
| point citations | 36 | **0** |
| range citations | 55 | n/a — each spans implementation |

So: **zero violations.** The wiki was already correct on this rule, and the reconciliation did not introduce anything. There is no debt to track and nothing for anyone to follow up on.

Apologies for the noise — the earlier comment should be read as withdrawn.

PR #1312, opened 2026-08-25
docs(backlog): close B-037 — lychee gate runs tokenless (customer view), re-measured clean

gmhelmold:
Closes the last open half of B-037 (the authenticated link checker).

**Re-measured, not assumed.** The gate already runs unauthenticated (no GITHUB_TOKEN in the docs-ci step or lychee.toml), and #1271 removed the private-repo evidence citations that produced the stale 2589 figure — zero `blob/tree/raw` links into `HumanGuardrail/corelink-server` survive in `apps/docs`. Tokenless re-run with the pinned `lychee 0.24.2`: github 21 OK / 1 error, humangr 20 OK / 1 error, both false positives (a `${ORG}` template + a POST-only newsletter endpoint 405). `docs-ci` broken-links (`fail: true`) is green on main without a credential.

`verify` becomes a real two-part check (no unschedulable runner label + no token in the broken-links job) so the item reopens if either regresses.


gmhelmold:
Heads-up from a parallel session: **B-037 is already closed on `main`** by #1310 (merged 2026-08-25T03:13Z), which is why this PR is now `CONFLICTING` — `BACKLOG.md` moved under it.

We reached the same verdict independently, so nothing here is wrong; the merged version just covers more ground and it's worth checking whether anything of yours is missing from it:

- **Full-corpus re-measurement**, not a per-host sample: the docs-build artifact from run `32801556993` re-checked with the same lychee 0.24.2, the same `apps/docs/lychee.toml` and the same `--remap` pair, from a host **outside Cloudflare** with `GITHUB_TOKEN`/`GH_TOKEN` unset. Byte-identical to CI — 108 870 total / 11 553 unique / 105 598 OK / **0 errors** / 3 183 excluded / 89 unsupported, differing only in redirects followed (111 vs 108). Your two "false positives" (the `${ORG}` template and the POST-only newsletter endpoint) come out **excluded** under the full config rather than counted, which is consistent with your read of them.
- The **vantage** half as well as the credential half: the job runs `runs-on: corelink`, i.e. from inside Cloudflare, checking links partly served by Cloudflare. That was the part that could not be settled by reading the workflow.
- A ⛔ banner on the lychee step, because lychee picks `GITHUB_TOKEN` up from the environment on its own — "we didn't pass one" has to be asserted, not assumed.
- The `verify` is **mutation-tested in three directions**, and notably exits **2** (`grep failed to look`) rather than passing when the workflow files are absent — the `!`-inverts-an-error trap that made B-042's first verify report CONFIRMED after `gh secret list` had simply errored.

Suggest closing this one unless you have something the merged item does not carry — happy to fold it in if so. Apologies for the collision; I'll check `gh pr list` by touched path before starting on a shared red next time.

gmhelmold:
Superseded — another session already closed B-037 on main with a stronger re-measurement (the full docs-build artifact re-checked tokenless off-Cloudflare: 108870 total, 0 errors, run 32801556993), and a verify that parses the broken-links job for any credential-shaped env at every level. My independent tokenless link-surface sample reached the same conclusion (0 customer-facing breaks; the 2589 figure was stale pre-#1271/#1297), so this PR is redundant.

PR #1316, opened 2026-08-25
perf(container): run the mandatory audit write concurrently with the R2 list call on native CAS/AC reads

gmhelmold:
## What

On the native CAS/AC `list()` read path, the mandatory `ListAttempted` audit
write (D1-over-HTTP INSERT) and the R2 `ListObjectsV2` enumeration used to run
STRICTLY SERIAL. They now run CONCURRENTLY under one `tokio::join!`, without
weakening fail-closed.

## Measured, live prod (colo=IAD, tenant d863fafb, 10/10 samples, full phase
attribution — post #1314)

Authed `GET /v1/ac/{tenant}?limit=1`:

```
ohop 54-60, ostore 61-96, oaudit 74-84, oother ~1, total 194-233 (ms)
```

The audit write (`oaudit`, 74-84ms) and the R2 list (`ostore`, 61-96ms) were
strictly serial — this change should collapse the pair to roughly the SLOWER
of the two (~85-96ms), i.e. shave something in the 60-95ms range off `total`.

## The concurrency seam

`crates/corelink-container/src/storage/r2_s3.rs` — `R2CasHandler::list` /
`R2AcHandler::list`: when the handler was built with the new async-audit seam
wired (`R2CasHandler::with_async_audit` / `R2AcHandler::with_async_audit`,
always true in production — see `build_r2_cas_handler_from_env` /
`build_r2_ac_handler_from_env`), the `ListAttempted` audit write and the R2
`ListObjectsV2` call are dispatched together via `tokio::join!` inside ONE
`block_in_place` + `block_on`, instead of two separate ones.

The audit half is a NEW narrow async method,
`D1AuditOutboxSink::append_async` (`crates/corelink-container/src/storage/d1_audit_sink.rs`)
— a bare `.await`-able future driving the SAME SQL/idempotency-key logic as
the existing sync `append`/`write_blocking` path (shared via a new
`build_insert` helper), but without `append`'s own `block_in_place`+`block_on`
wrapper, so it composes into `tokio::join!`. The `AuditSink` trait itself
stays SYNC and untouched — `append_async` is a plain inherent method on the
CONCRETE `D1AuditOutboxSink`, reached because `R2CasHandler`/`R2AcHandler` now
also hold `Option<Arc<D1AuditOutboxSink>>` alongside the existing
`Arc<dyn AuditSink>` (same instance, two views — the production builders wire
one `Arc` into both). A handler without that seam (every test handler in this
module) keeps the fully serial fallback, byte-identical to before this PR.

### Why this cannot serve bytes on a failed audit

The two calls are DISPATCHED together, but the RESPONSE is still gated on
BOTH results, checked in the SAME order the old serial code checked them:
the audit result first — `Err` short-circuits to `CasHandlerError::AuditFailed`
/ `AcHandlerError::AuditFailed` exactly as before — and only if that succeeds
is the store result inspected and rows returned. So:

- **"No bytes served without the audit row" — preserved exactly.** A failed
  audit write still returns 503 `AuditFailed`; no R2 rows ever reach the
  caller.
- **"No R2 request issued without the audit row" — relaxed, on purpose.**
  This was never a stated invariant on its own (only "audit before
  enumeration" was, meaning "before the response", which still holds); it is
  the guarantee traded away for the latency win. Concurrency means an R2
  `ListObjectsV2` call can now be dispatched even when the audit write is
  about to fail — that request's result is simply discarded.
- **Cross-tenant denial (`ListDenied`) is untouched** — it returns before
  reaching the concurrent block and never joins anything; it still cannot
  issue an R2 call at all.

## Which paths I left serial, and why

- **`R2CasHandler::read` / `exists`, `R2AcHandler::lookup`** — same
  "audit-then-object-call" shape, but with `resolve_byok` + key derivation
  genuinely in between (not just a no-op today, but a real intermediate step
  the R2 key depends on). Parallelizing these safely needs the same
  audit-future / prepare-future `tokio::join!` pattern applied to a 3-step
  chain, which is a materially bigger, separately-reviewable change. Left
  serial this PR; only `list()` (the profiled endpoint) went concurrent.
- **`write` / `update` / `delete` (every mutation path)** — left FULLY
  SERIAL, deliberately. The `WriteAttempted`/`DeleteAttempted` audit write
  gates a MUTATION (a PUT/DELETE), not a read: if that audit write and the
  mutation ran concurrently and the audit failed, the mutation could still
  have already landed — an R2 write/delete succeeding when its audit row
  never did. That is exactly the guarantee the brief said must never be
  weakened, so these paths are unchanged.

## `origin_timing` — the arithmetic stays honest

`crates/corelink-container/src/origin_timing.rs` gets a new "Concurrent
native-plane list seam" doc section. On the concurrent `list()` path, `oaudit`
and `ostore` now genuinely overlap in wall-clock time — if both individually
opened their own `PhaseScope`, `Σ(phases)` could exceed `total` and the
existing `oother = (total - Σ).max(0)` guard would silently swallow the
overcount (still never negative, but no longer an honest partition).

Fix: the whole joined window is timed ONCE, under `Phase::Store` only.
`append_async` (the audit half of the join) deliberately never opens its own
`PhaseScope` — only the fully-serial `write_blocking` bridge (used by every
OTHER audit call, including every mutation path and `list()`'s own serial
fallback) does. So on the concurrent `list()` path: `ostore` reports the full
audit+R2 window, `oaudit` reports nothing for that specific write, and the
four-way sum still partitions the request exactly — by construction, not by
the `max(0)` guard papering over a double-count. `oaudit` is unaffected
everywhere else.

Also reconciled: `docs/knowledge/compliance/audit-chain.md`,
`docs/knowledge/planes/request-flow.md`, `docs/knowledge/storage/r2-cas-bucket.md`
(cited line ranges + the new concurrent-seam behavior).

## Gates (verbatim results below in the PR discussion / commit)

- `cargo test -p corelink-server` (unfiltered)
- `cargo clippy --all-targets -- -D warnings`
- `cargo fmt --all --check`
- `python3 scripts/validate_okf.py`
- `python3 scripts/validate_specs.py`


gmhelmold:
Cold review — merging. Two notes for the record, neither blocking:

**1. A flaky-under-load test, unrelated to this change.** Running the full suite on this branch while the Mac was also compiling another cargo job, `routes::cas_erase::tests::bloom_concurrent_reads_and_writes` failed with `written digest digest-0002 must be tombstoned` (1377 passed / 1 failed). Re-run in isolation it passes **5/5**, including with `--test-threads=4`, and the unloaded full run on this same branch passed 1378/0. It touches `cas_erase`, nothing this PR modifies. So: load-sensitive test, not a regression here — but it is a real timing dependency worth knowing about the next time CI reports it.

**2. The overlap is the part I checked hardest, and it holds.** `append_async` opens no `PhaseScope`, so the joined window is attributed once to `Phase::Store` and `Σ(phases) ≤ total` stays true *by construction* rather than by `oother`'s `max(0)` guard absorbing a double count. The fail-closed argument is also the right one: concurrency changes whether the R2 call is *dispatched*, never whether its result can *reach the caller* — the audit result is still checked first, in the same order, and cross-tenant denial still returns before the join.

Expected effect, against the measurement this targets (`colo=IAD`, 10/10 samples): `oaudit` 74-84 ms and `ostore` 61-96 ms were strictly serial; the pair should now cost roughly the slower of the two. I will re-measure from inside the fabric after the image rolls and post the numbers here.

PR #1322, opened 2026-08-25
ci(e2e): wire real-browser money-path suite on self-hosted fleet (B-042)

gmhelmold:
## What
Wires `tests/e2e-browser` — the real-browser Playwright + @clerk/testing suite — so the money path can be exercised in CI. Its `03-money-checkout` spec is the only end-to-end proof a customer reaches a live `checkout.stripe.com/…/cs_live_` session (stops at the hosted page; no card submitted, no charge).

## How
- New `.github/workflows/e2e-browser-prod.yml`: `workflow_dispatch`-only, runs on `[self-hosted, mac, corelink-builder]` (has a real browser; `playwright install chromium` works) — **zero GitHub-hosted spend**.
- Standalone `npm ci` (the suite carries its own `package-lock.json`); deliberately **not** a `pnpm-workspace.yaml` member — its Playwright deps don't belong in the root pnpm graph, and the workflow installs them itself, which is what the orphan actually needed.
- Owner authorised the credential: `CLERK_LIVE_PUBLISHABLE_KEY` + `CLERK_LIVE_SECRET_KEY` are now repo secrets (set from `.env.local`, verified by name).

## Why dispatch-only
Each run mints a live prod Clerk user + a real (unpaid, expiring) Stripe session; the fixture DSR-deletes the user on teardown. Promoting to a nightly `schedule:` is a deliberate separate decision, deferred.

Closes **B-042**.


gmhelmold:
Superseded: #1298 already adopted the browser suite on main (corelink/pnpm workflow + pnpm-workspace membership), a better version than this mac/npm duplicate. The only genuinely-missing piece was the two live Clerk credentials — CLERK_LIVE_PUBLISHABLE_KEY + CLERK_LIVE_SECRET_KEY are now repo secrets (owner-authorised, set from .env.local, verified by name), matching exactly what #1298's workflow reads. B-042 is complete without this PR.

PR #1343, opened 2026-08-26
fix(billing): guard the tier upsert against resurrecting a canceled subscription

gmhelmold:
## The bug

`SQL_UPSERT_TIER` wrote `subscription_state='active'` with **no condition on the tenant's current billing status**. The payload guard `subscription_status_grants_access` (`handler.rs:141`) only inspects the event body — so a **stale, out-of-order** `customer.subscription.updated(status=active)`, legitimately signed and simply redelivered late by Stripe, passed it and re-granted paid access **after** the cancel had already been materialized.

The twin writer already closed this at DB level:

```ts
// apps/signup-worker/src/webhooks/stripe.ts:863
AND status != 'canceled'
```

Its comment at `:833` names the scenario by hand. Two writers of the same access gate with opposite postures is the drift class this repo has been bitten by before.

## The fix

The `DO UPDATE` arm now carries:

```sql
WHERE NOT EXISTS (
  SELECT 1 FROM tenant_billing tb
   WHERE tb.tenant_id = tier_selections.tenant_id
     AND tb.status = 'canceled')
```

The subquery is **correlated, not parameterized** — zero extra placeholders, so the arity stays at exactly 4 and no call site changes. An existing test pins that count.

## Proven before the edit

Executed against the real DDL of `0039_tier_selection.sql` + `0055_tenant_billing.sql`:

| case | result |
|---|---|
| `canceled` + existing row | **not** resurrected |
| `paid` + existing row | granted |
| no `tenant_billing` row | granted — first purchase never blocked |
| placeholders | 4 |

`SQL_DOWNGRADE_TIER` untouched (pinned by a new test).

## Known residual — documented, not hidden

The guard covers the `DO UPDATE` arm only. A tenant with `status='canceled'` and **no** `tier_selections` row still gets a fresh active row. Safe: the cancel path writes `'inactive'` rather than deleting, so a missing row means *never had a tier*. The rustdoc also records why closing it via `INSERT..SELECT..WHERE NOT EXISTS` is the wrong move — it breaks the `ON CONFLICT` interplay and the bind arity.

## Verification

```
cargo test -p corelink-billing-stripe-materializer   56 passed, 0 failed  (21 -> 24 #[test])
validate_specs.py                                    484 total, 0 failures
validate_secrets_matrix.py                           code_only=0, exit 0
check_changelog_duplicates.py                        no duplicates
validate_okf.py                                      161 concepts, 0 stale, 0 drift
```

OKF reconciled in a **separate commit** after the code commit, per the C5 rule.


gmhelmold:
Superseded by #1347, which lands the byte-identical `SQL_UPSERT_TIER` guard with a strictly stronger test: #1347 pins the statement behaviorally against the REAL 0039+0055 DDL in an in-memory SQLite (`rusqlite`, bundled), where this PR only asserted on the SQL string. #1347 is also the base of a five-PR stack (#1348–#1352), so keeping it avoids retargeting that chain. Closing this one as the duplicate; no unique coverage is lost — `downgrade_tier_sql_unchanged` here is covered by `tier_sql_unchanged_and_schema_correct_0039` there.

PR #1350, opened 2026-08-26
fix(billing): DLQ de webhook duravel em D1 + evidencia de auditoria duravel (MED-5)

gmhelmold:
> **Correction (post-merge).** This body says migrations `0094`/`0095`. Both ordinals were already taken on `main` (`0094_runner_usage_counter.sql`, `0095_usage_event_staging_aggregatable.sql`), so the files shipped as **`0104_webhook_dlq_event_id_unique.sql`** and **`0105_stripe_billing_audit_events.sql`**. #1353 added an ordinal-uniqueness check to `check_migrations_additive.py` so a reused number now fails CI. New migrations start at `0106`.

---

**WP-D1 (P1) — DLQ de webhook duravel em D1 + evidencia de auditoria duravel (MED-5).**

A quarentena do dispatcher era InMemory: cada redeploy destruia a UNICA copia do evento (dedup-before-materialize F-008 e deliberado; retries do Stripe batem em AlreadyProcessed). D1WebhookDlqStore implementa WebhookDlqStore sobre stripe_webhook_events_dlq (0045 + UNIQUE(event_id) nova 0094 — UPSERT SQL impoe DLQ_IDEMPOTENT_ON_EVENT_ID). MED-5 fechado no mesmo PR: D1BillingAuditEmitter anexa audit-before-mutation em stripe_billing_audit_events (0095, RETAIN no registro DSR). Um unico D1HttpClient compartilhado pelos 3 consumidores; dev/CI mantem in-memory.

Desvios documentados: impls no container (direcao de dependencia, precedente billing_d1_http.rs); adapter_d1.rs recebeu as tabelas novas no registro CF-1. 2 failures de checkout_promo sao pre-existentes em main (dependem da API live Stripe).

Stacked em #1349.


gmhelmold:
Blocker on this PR, found while triaging merge order — **both migrations reuse ordinals that already exist on `main`:**

```
main:      migrations/d1/0094_runner_usage_counter.sql
this PR:   migrations/d1/0094_webhook_dlq_event_id_unique.sql

main:      migrations/d1/0095_usage_event_staging_aggregatable.sql
this PR:   migrations/d1/0095_stripe_billing_audit_events.sql
```

CI is green on this because `check_migrations_additive.py` only scanned for destructive tokens — it never checked ordinal uniqueness. Two sessions branching off the same `main` each compute "next = 0094" independently, so this is the default outcome of parallel work rather than a slip.

It matters for two reasons, both silent: apply order between two same-ordinal files degrades to a lexicographic tiebreak on the rest of the filename, and the D1 ledger keys on the full filename so both record as applied and nothing errors. The cost surfaces during a replay or an incident.

**Fix here:** renumber to the next free ordinals. `main` currently ends at `0102`, and `0103` is claimed by #1344 (`0103_audit_outbox_sealed_tail_index.sql`), so this PR should use **`0104`** and **`0105`**. These files have not been applied anywhere yet, so renaming them is free — unlike the `0044` pair from 2026-05-14, which is already in production and cannot be renamed without desyncing the ledger.

#1353 adds the missing gate so this cannot recur; once it lands, this PR will go red until the files are renumbered.

PR #1359, opened 2026-08-26
feat(audit): emit the ReadAttempted rows an edge-served findMissingBlobs owes

gmhelmold:
Implements the seam [#1355](https://github.com/HuGR-Labs/corelink-server/pull/1355) specified (B-047). Container half only — the edge call is the next PR, so nothing changes behaviour yet.

## Why

F1 (#1340) measured the edge answer at **43/43 parity, 8.65 s → 2.10 s at n=100**, which is why F2 looked like a flag flip. It is not. `probeMissingAtEdge` computes the same SET; `exists_batch` (`crates/corelink-container/src/storage/r2_s3.rs:1560`) does four things *around* that computation which the edge does not:

1. one `ReadAttempted` row per digest into `audit_outbox`;
2. `ReadDenied` audited **before any dispatch** on a cross-tenant digest;
3. the fail-closed coupling — probe results already in hand are discarded when the audit write fails;
4. the `AvailCasGet` / `LatencyCasGetP99` SLIs.

Serving from the edge today answers a REAPI **read** surface with its evidence absent, and keeps answering while the audit sink is down.

## What

`POST /_internal/audit/cas-attempted`. The edge probes R2 through its in-colo binding and **awaits** one call here, which emits the rows through the same `D1AuditOutboxSink` the container already uses. `204` is the only outcome that authorizes the caller to serve; anything else means fall through to the container, which attempts the audit itself and fails closed in its own taxonomy.

Writing `audit_outbox` rows from the Worker was rejected: UUIDv7 `id`, CloudEvents `payload_json`, `UNIQUE (request_id, event_type)` and a region trigger that `RAISE(ABORT)`s make that row a contract, and a second TypeScript author would drift while still passing its own tests.

- **Auth:** dedicated `CORELINK_AUDIT_ATTEMPTED_AUTH_KEY`, **no** shared-key fallback (the `/_internal/pat/mint` posture) — it writes tenant-attributed rows for an arbitrary tenant. Unset/<32 chars ⇒ unmounted ⇒ the edge's call 404s and it falls through. Never an outage, only the slow path.
- **Cross-tenant:** refused, but audited as `ReadDenied` first.
- **Cap:** 256, mirroring the edge's own subrequest-budget cap; asserted to stay under `FIND_MISSING_BLOB_CAP`.

## Gates run locally

- `cargo check -p corelink-server` — clean (exit read directly, not through a pipe)
- `cargo test -p corelink-server audit_cas_attempted` — 3 passed, 0 failed
- `cargo fmt --all --check` — clean
- `validate_okf` — 161 concepts, 0 stale, 0 drift (10 concepts reconciled: the mount shifts every cited range below it; `planes/container` also claimed the archive route is "mounted immediately after" the drain, which this change makes false — re-authored, not renumbered)
- `secrets-checklist-verify` OK, `validate_secrets_matrix` code_only=0
- `validate_specs` 484/484, `backlog_verify` 0 drifted


gmhelmold:
Not splitting this one, and the number is misleading about why.

Of the 620 insertions, **246 are the OKF reconcile across 18 concept files** (mostly `+1/-1` citation renumbering — the diff counts each as an insertion plus a deletion). The actual new code is one route module, its mount, and two lines of env wiring.

The reconcile cannot be a separate PR: `okf-wiki-validation` is a **blocking per-PR gate**, and the drift is *caused by* this PR's insertions (adding a route module shifts every cited line below it in `routes.rs` and `main.rs`). A code-only PR would be red on arrival, and a docs-only follow-up would have nothing to reconcile against until the code landed.

The one concern is intact: the container half of the F2 audit seam (B-047). The edge-side awaited call is a separate PR, by design.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #1403, opened 2026-08-28
fix(devenv): apply cargo fmt --all to fix rustfmt CI gate

gmhelmold:
Applies cargo fmt --all to corelink-container/src/main.rs to fix the rustfmt check that fails in CI. The change is cosmetic-only (one expression reformatted from 1 line to 4 lines) but is needed to keep the gates green as more devenv work lands.


gmhelmold:
Cannot pass CI: only 1 of 5 self-hosted mac runners is online (corelink-builder-2), and historical commits lack Signed-off-by (DCO). Closing for now. The fmt fix D13 is on pr-0a-d13-rustfmt branch. Re-apply with git rebase --signoff or use admin-merge.

PR #1411, opened 2026-08-29
ci(rustfmt): also run on push to main, so a red main cannot stay invisible

gmhelmold:
`rustfmt.yml` triggers on **`pull_request` only**, so nothing checks `main` after a merge. A red main stays invisible until the next author inherits a gate they did not break.

Not theoretical: **#1406 — titled *"apply cargo fmt --all to fix rustfmt CI gate"* — merged the inverse** of correct formatting, while red, and `main` sat rustfmt-red with nothing to report it.

Adds `push: branches: [main]` on the same `paths:` filter and the same self-hosted runner.

No tension with CLAUDE.md's cron rule ("a cron earns its keep only when something can change WITHOUT a commit"). This is not a clock — formatting changes only when code changes, so it is the push half of an existing code-change trigger, and costs nothing extra per merge. It runs post-merge on the self-hosted fabric, `required checks = []`, and PR/push lanes do not share a concurrency group, so it cannot block `main`.

## Scope reduced — one file

This PR originally also carried the formatting fix and its OKF re-anchor. Both are dropped:

- **#1410 already deletes the `adapter_d1.rs:97` offender outright** (it is the phantom-table line) and carries a byte-identical `main.rs` hunk. `cargo fmt --all --check` already passes there.
- The two PRs **conflicted on `checkpoint_sha` in six OKF concepts** — `git merge-tree` reports 7 files. A conflicting PR gets **zero checks**, which is exactly how a gate reads green over unverified work.

What remains is the part #1410 does *not* do: closing the hole that let #1406 through. One file, no overlap, no conflict.


gmhelmold:

Gated at `8b5b091a7f65af5e484d601e7a74f457181c2598`, merged by `scripts/pre-merge-gate-check.sh --merge 1411`.

```
[pass]     Verify all GHA actions SHA-pinned
[pass]     actionlint (native, SHA-pinned tool)
[pass]     dco
[pass]     gitleaks detect
```

PR #1418, opened 2026-08-30
fix(ci): arm the TypeScript coverage floors — they had never once executed

gmhelmold:
## WP-3 — enforce a coverage floor so coverage cannot silently rot

Part of the remediation roadmap (`docs/campaigns/remediation/ROADMAP.md`, #1414). Inherited from the audit handoff (item 13).

### The headline finding: the floor already existed and had never once run

`worker/vitest.config.mts` has carried a full `coverage.thresholds` block since the file was written. `.github/workflows/worker-vitest.yml` ran `npx vitest run` — **with no `--coverage`**. Vitest evaluates `thresholds` *only* when the coverage provider is enabled, so the block was inert config, not a gate.

This is the fifth gate this campaign has found that structurally could not see what it existed to catch.

**It had already cost us.** Measured on `origin/main`, `src/index.ts` is written down as 90% lines and actually sits at **77.13%**. It rotted 13 points and nothing went red, because nothing was ever measured.

### Measured current vs chosen floor

Measured 2026-08-29 on `origin/main`, from real runs (not estimates).

| Target | Metric | **Measured** | **Floor** | Slack |
|---|---|---|---|---|
| `worker/` total | lines | 84.97% | **80** | 4.97 |
| `worker/` total | statements / functions / branches | 84.01 / 90.55 / 78.14 | 79 / 85 / 73 | ~5 |
| `worker/src/index.ts` | lines | 77.13% | **72** | 5.13 |
| `worker/src/durable_object.ts` | lines | 81.69% | **76** (was 35) | 5.69 |
| `worker/src/rollout_controller.ts` | lines | 100% | **90** (was 0) | 10 |
| `apps/signup-worker/` | lines | 79.09% | **74** | 5.09 |
| `apps/signup-worker/` | statements / functions / branches | 77.81 / 76.87 / 67.47 | 72 / 71 / 62 | ~5 |

Every floor is **below measured, with slack**. None is aspirational. This org already carries five workflows with 1,057 lifetime runs and zero successes; a floor above current coverage would join them and train everyone to ignore CI. The floor's job is to catch regression, not to aspire.

### Net effect on rigor: strictly up

Two per-file thresholds are **tightened**, because they were too loose to catch anything:

- `durable_object.ts` **35 → 76**. The old comment claimed ~41% was the ceiling because the CF Container API is unreachable from Node.js. It measures 81.69% today.
- `rollout_controller.ts` **0 → 90**. A `0` threshold cannot fail. The Phase B stub is in fact 100% covered.
- Global `lines` **65 → 80**.

One is **lowered**: `index.ts` 90 → 72, to match the measured 77.13%. This is not weakening a gate — that number has been enforced zero times in its life. Before this PR it is unenforced and unmeetable; after, it is enforced on every PR. **The gap is recorded test debt, to be repaid by ADDING tests, never by editing the number again.** No test was deleted or weakened.

### The second hole: `coverage.include` was implicit

Both v8 and istanbul default to reporting only files a test *happened to import*. Under that default, **adding a brand-new untested source file leaves the percentage untouched** and the floor is blind to it.

Both configs now set `include: ["src/**/*.ts"]` explicitly. Measured on signup-worker: **88.23%** (imported-files-only) vs **79.09%** (all of `src/`) — ~9 points of code the default could not see.

### Proof (bare exit codes, never through a pipe)

`cmd | tail` returns tail's status; that mistake let a PR merge here with 4 checks pending. Every number below is a bare `$?`.

| Check | Result |
|---|---|
| `worker`: passes at chosen floor | **exit 0** |
| `worker`: fails when floor breached (global lines → 99) | **exit 1**, `ERROR: Coverage for lines (84.97%) does not meet global threshold (99%)` |
| `signup-worker`: passes at chosen floor | **exit 0** |
| `signup-worker`: fails when floor breached (lines → 99) | **exit 1** |
| `signup-worker`: new untested file moves the total | 79.09% → **78.86%** (denominator 1019 → 1022) |
| `signup-worker`: clean `rm -rf node_modules && npm install --legacy-peer-deps` then `--coverage` | **exit 0** |

### What would break this and NOT be caught

Answered honestly, per the campaign rule:

1. **Deleting `--coverage` from the workflow step.** This is the single point of failure and it fails *silently* — the suite still runs, still goes green, and the floor evaporates. This is exactly how the `worker/` floor sat dormant. **Mitigated only by comments** at both the workflow step and the config block stating the flag is load-bearing and paired. A structural fix (a meta-check asserting every config with `thresholds` is invoked with `--coverage`) is the honest follow-up; I did not build it here.
2. **A small untested file.** The gate *sees* new untested code (proven above), but the ~5-point slack is a tolerance budget: a 3-line untested file moved the total only 0.23 points and would not trip the floor. That is deliberate — zero slack means a flaky-red gate — but it means rot below the slack is invisible until it accumulates.
3. **Coverage that is high but meaningless.** Tests that execute lines without asserting on them satisfy this gate completely. A line-coverage floor measures execution, not verification. Mutation testing is the answer to that, and it is out of scope here.
4. **Path-filtered lanes.** `worker-vitest.yml` fires only on `paths: worker/**` and signup-worker likewise. A change elsewhere that reduces those packages' coverage would not run the gate. Both floors only defend their own package.
5. **The other four TS packages carry no floor at all** (see below).

### Scope: what is deliberately NOT covered

- **Rust (`corelink-server`).** `coverage.yml` is **`workflow_dispatch`-only** — no `pull_request`, no `push`, and its weekly cron is commented out in-file. Per the frozen contract, if no coverage lane runs on PRs, **do not create one**: two full instrumented workspace builds on the owner's Mac (the self-hosted runners) do not belong on a PR. **No Rust floor added, by design.**
- **`apps/admin-ui`, `apps/docs`, `apps/analytics-worker`, `apps/get-corelink-worker`.** Measuring these requires a full pnpm-workspace install. Per the standing rule *"if you cannot measure a repo's current coverage cheaply, set no floor there and say so — do not guess"*, **no floor was set and no number was guessed.** Named here rather than passed over in silence.

### Companion PRs

- `corelink-runners`: spawn-Worker gate — measured 80.97% lines, floor 75. (See its PR.)
- `corelink-workspaces`: **already done, nothing to do.** `.github/workflows/coverage.yml` runs on `pull_request` + `push`, on `runs-on: corelink`, and already enforces `FLOOR=88` with a real `exit 1`. Verified current = **90.10% lines**, stable byte-identical across the last 4 runs (32791835190, 32776068695, 32614038126, 32412470673). 2.1 points of slack; left untouched rather than ratcheted, to avoid narrowing a working gate for marginal gain.

### Runners

No new hosted runners. Both touched lanes stay on `runs-on: corelink`.

### Note on the changelog

Rebased onto `a2f41082`. This PR uses the `changelog.d/` **fragment** mechanism landed in #1417 rather than editing the shared `CHANGELOG.md`, so it cannot conflict with the rest of the in-flight queue. Verified: `python3 scripts/assemble_changelog.py --lint` → **exit 0**, `OK 1418-coverage-floor.md [Fixed]`.

(The first push of this branch did edit `CHANGELOG.md` and went `CONFLICTING` — which produced **zero `pull_request` runs**, leaving only the `pull_request_target` metadata jobs reporting SUCCESS. That is the documented trap where the most dangerous PR state produces the most reassuring output; it was caught by checking `mergeable`, not by reading the green checks.)


gmhelmold:
### ⚠️ `published SDK artifacts match their source` is RED — and it is **pre-existing on `main`**, not caused by this PR

Verified the way this campaign requires — on a **clean `origin/main` worktree**, not on this branch:

```
git worktree add --detach /tmp/wt-main-check origin/main   # HEAD = a2f41082
bash scripts/verify-sdk-artifacts.sh; echo $?             # -> 1   (bare exit code)
```

`origin/main` at `a2f41082` fails the identical check with the **identical diff**. The drift is in the published JS SDK tarball under `apps/docs/static/`, which is stale against `sdks/js` source:

```
- const res = await this.rawRequest("GET",  this.casPath(digest));
+ const res = await this.rawRequest("HEAD", this.casPath(digest));
```

(plus the matching doc-comment and `client.d.ts` drift). The Python wheel and the Go zip both report `ok`; only the npm tarball is stale. Someone changed `stat()` from a full `GET` to a `HEAD` in the SDK source and did not republish the tarball that `apps/docs/static/` serves to customers.

**Nothing in this PR touches `sdks/js`, `apps/docs/static/`, or any SDK build input.** This PR's files are two vitest configs, two workflow steps, a package.json, a changelog fragment, and `pnpm-lock.yaml`.

Why it shows as absent on most other open PRs: the job is path-filtered, so it simply does not run on them. #1416 passes it because it is based on an older `main`, from before the SDK source change landed.

**This is a real defect and it is not mine to fix inside a coverage-floor PR** — the remedy is republishing the tarball per `docs/internal/sdk-publishing.md`. Flagging for the lead: `main` is currently red on this gate, and customers are being served an npm tarball whose `stat()` still does a full-body `GET` instead of a `HEAD`.

gmhelmold:
**Nota de honestidade sobre este merge.** Ele foi feito pelo `pre-merge-gate-check.sh --merge 1418` — mas de um worktree que está no meu branch de trabalho, e portanto de uma **cópia do script anterior ao #1428**. O portão gateou corretamente (todas as lanes verdes, listadas na saída), mas **não deixou o registro do conjunto de lanes** que o #1428 tornou obrigatório: aquela cópia não tem o mecanismo.

Não vou escrever o registro à mão aqui. O valor do #1428 é justamente o registro ser um sub-produto do gating e não algo que alguém redige depois; um comentário hand-written com a mesma cara seria uma falsificação da garantia.

O que cabia era tornar o erro impossível: [#1429](https://github.com/HuGR-Labs/corelink-server/pull/1429) faz o portão comparar o próprio blob com o de `origin/main` e **recusar o merge** quando não é a mesma versão.

PR #1428, opened 2026-08-30
fix(gate): record the checked lane set on the PR as a by-product of gating

gmhelmold:
## The defect

WP-1 requires every merge to record which checks authorised it. Measured across nine consecutive merges: **zero did.**

The criterion was not too strict. The mechanism was that the operator had to remember — and remembering is not a mechanism. Lowering the criterion to match the observed behaviour would have been the wrong repair; this removes the human step instead.

## What changes

`scripts/pre-merge-gate-check.sh --merge` now writes its lane set at the one point it concludes green, and posts it to the PR once GitHub has confirmed `state=MERGED`.

Two properties follow from *where* the write sits, rather than from anyone's care:

- a red gate produces **no record at all** — the write is downstream of the green verdict;
- the record renders from the **same list the gate decided on**, so it cannot claim a greener check set than the one that actually authorised the merge.

An `--admin` override is carried into the comment along with the caller's stated reason, so an overridden merge cannot look like a clean one in the record.

Posting is best-effort and last. This script's exit status answers exactly one question — *"is PR #N merged"* (#1051) — so a failed comment warns loudly rather than failing, because a silent failure here would rot the record back into the state this fixes.

## Verification

Run against real check data, not fixtures:

| case | data | result |
|---|---|---|
| green rollup | #1427's actual checks | writes all 10 lanes, buckets intact |
| red rollup | #1418's actual checks | writes **nothing** |
| `--dry-run` | #1418 | unchanged, still refuses |

The end-to-end post is proven by **merging this PR with the mechanism it adds** — the lane-set comment on this PR is the test result.


gmhelmold:

Gated at `4921705e2f7d75788cb6c741f914db62e65d7e92`, merged by `scripts/pre-merge-gate-check.sh --merge 1428`.

```
[pass]     Top-level python test suite
[pass]     dco
[pass]     gitleaks detect
```

PR #1432, opened 2026-08-30
feat(devenv): re-cut the DevEnv customer surface from the frozen manifest (PKG-3)

gmhelmold:
The DevEnv feature was rejected wholesale for a re-cut. This brings back only what `docs/campaigns/remediation/devenv-manifest.tsv` marks `SALVAR` without a `LOTE SEPARADO` marker.

## Completeness predicates — counted, not trusted

**1. Manifest tally.**
```
wc -l           : 174
data rows (NR>1): 173
  59 ARQUIVAR
  98 DESCARTAR
  16 SALVAR
```

**2. The 6 `LOTE SEPARADO` rows, named before exclusion** — all basePath/pricing, none is the feature:
`apps/admin-ui/src/app/llms.txt/route.ts` · `.../pricing.txt/route.ts` · `.../pricing/route.ts` · `.../lib/public-url.ts` · `.../tests/pricing-page.test.tsx` · `.../tests/pricing-routes.test.ts`

**4. All 10 in-scope rows present** (16 − 6), each verified against the working tree.

## 3. The migration's two blocking conditions

**Renumbered 0094 → 0106.** `0094` on `main` is `runner_usage_counter`; a duplicate ordinal fails `check_migrations_additive.py`. The ordinal was allocated against `origin/main` **and every open PR at push time**, not when this was written.

**`devenv_monthly_vcpu` registered in `TENANT_ID_TABLES` and `ALL_TENANT_KEYED_TABLES` in this same PR**, classified ERASE per ADR-S11-013 (pre-invoice usage aggregate, same class as `runner_usage_counter`; the fiscal record is the retained Stripe row).

This is the half that matters. #1405 put this name in both registries citing a migration that creates a *different* table; the phantom split an Art.17 erasure in half — 16 operational tables destroyed, `byok_*`/`tenant` never reached, 500 with no attestation. Shipping the table **without** the registry entry inverts that into silent **under-erasure**: no 500, no signal, PII quietly retained. Both CF-1 mirror gates pass:

```
every_registry_table_is_actually_created_by_a_migration ... ok
every_migrated_tenant_keyed_table_is_classified ... ok
```

## `wrangler.toml` is hunk-level

Only the `RUNNER_DEVENV_DO` binding — added top-level **and mirrored into `[[env.prod.durable_objects.bindings]]`**. `durable_objects` is non-inheritable: an env block *replaces* the top-level list. The top-level hunk alone ships DevEnv green in dev and permanently 503 in prod, with CustomerNav linking users at it. Verified by parsing the TOML, not by reading it: **7 bindings in both blocks, zero unmirrored, no crons added.**

## Deliberately not taken

The six prod crons (three top-level plus three in a new `[env.prod.triggers]`), the `AUDIT_BUCKET` binding, the plaintext Ed25519 seed as a non-secret prod var, `CORELINK_API_BASE` (nothing under `worker/src` reads it — checked), and an unrelated `undefined`→`null` hunk in the replication-coordinator body.

That last one is worth a note: it was fixing a **real pre-existing type error**. Dropping it is still right for a re-cut, and it costs nothing — `tsc` reports **17 errors on `main` and 17 here**, and a set-diff (not a count-compare) shows the only difference is `index.ts(2176)` → `index.ts(2209)`, the same error shifted by this PR's additions. Zero new. The drive-by fix is legitimate and belongs in its own PR.

## OKF: a new concept, not an exclude

Touching `index.ts`, `wrangler.toml` and `adapter_d1.rs` shifted **133 citations**; all remapped from the real per-concept hunks (the remapper refuses if a citation lands inside a rewritten region — none did), with blob anchors re-pinned to the committed blobs.

The two new worker files tripped C10b. `openapi_devenv.ts` is a static document and would have excluded cleanly, but **`devenv_guard.ts` is a request-reachable enforcer** — it reads `runners_entitlement` and returns allow/deny on the devenv path. The manifest's own rule is that a load-bearing edge file with no concept must RED the gate, so excluding it would have been silencing, not satisfying. `surfaces/devenv-edge` is a real concept with 8 citations, each opened and checked against the code.

`validate_okf.py`: **165 concepts, 0 stale, 0 drift**, exit 0.

## Gates (exit codes read directly, no pipe)

`clippy -p corelink-server --all-targets -- -D warnings` **0** · `cargo fmt --all --check` **0** · `validate_specs` **0** · `validate_okf` **0** · `check_migrations_additive` **0** · `validate_secrets_matrix` **0** · `secrets-checklist-verify` **0** · `check-env-contract` **0** · `tsc` no new errors.

## One gate I did not touch, on purpose

`backlog_verify` is **red on B-061**, and it is not this PR's content. B-061 asserts a fixed count of orphaned OKF anchors and goes DRIFTED *"nos dois sentidos: se alguém reancorar"* — so any PR that re-anchors trips it, including the work that fixes what it describes. Measured: **59 on `origin/main`, 55 on this branch**, while the verify demands **63**. It is another session's item and the number will change again at squash, so I am reporting it rather than editing it.


gmhelmold:

Gated at `a00e9d7640d74dfd45ccc022b42a0eed26c3ea3f`, merged by `scripts/pre-merge-gate-check.sh --merge 1432`.

```
[pass]     Cross-validate INV registry × TLA × code × tests
[pass]     D1 migration validate
[pass]     PR gate (crate-scoped clippy + tests + canonical vectors)
[pass]     PR gate (crate-scoped clippy + tests + canonical vectors)
[pass]     PR gate (crate-scoped clippy + tests)
[pass]     WASM target build (CF Workers compile compatibility)
[pass]     build
[skipping] cargo-fuzz nightly (1h each)
[skipping] cargo-fuzz nightly (1h)
[pass]     cargo-fuzz smoke (60s)
[pass]     cargo-fuzz smoke (60s, both targets)
[pass]     cargo-mutants (changed lines only)
[skipping] cargo-mutants kill-rate ≥ 80%
[skipping] cargo-mutants kill-rate ≥ 80%
[pass]     cf:build
[pass]     dco
[pass]     gitleaks detect
[pass]     lint
[pass]     playwright critical-flows (chromium)
[pass]     playwright legacy (chromium)
[pass]     proptest density gate
[pass]     radar
[pass]     test
[pass]     typecheck
[pass]     vitest
```

PR #1439, opened 2026-08-30
fix(admin-ui): the public pricing redirects pointed at 404s; plus the null-body hunk

gmhelmold:
Packages **A** and **B**, same source (the rejected #1397), one PR because they share a review.

## A — the six `LOTE SEPARADO` rows

Held out of the PKG-3 devenv re-cut on purpose: a customer-facing redirect reviewed inside a PR titled *"devenv"* gets read by someone looking at something else. All six are new files, **106 lines**, none of them the devenv feature.

### They do not ship as written

The three routes redirect to `https://humangr.com/corelink/{pricing,pricing.txt,llms.txt}`. **All three 404 on the live site.** The apex equivalents return 200:

| URL | status | | URL | status |
|---|---|---|---|---|
| `/corelink` | **200** | | `/corelink/pricing` | **404** |
| `/pricing` | **200** | | `/corelink/pricing.txt` | **404** |
| `/pricing.txt` | **200** | | `/corelink/llms.txt` | **404** |
| `/llms.txt` | **200** | | | |

The section root *is* under `/corelink`; the pricing page and the two text endpoints live at the **apex**. Prefixing the latter produces a **307 into a 404 on the buyer funnel** — invisible to any build or type check.

### The tests locked the defect in

The tests that shipped with these routes asserted the **404 URLs** as expected values. They passed while the behaviour was broken.

Asserting a string is exactly what allowed that, so the guard added here asserts **the shape that was wrong** — no public destination may carry the `/corelink` basePath — rather than a second string that could be wrong the same way.

`public-url.ts` shipped **unused** (the routes hardcoded their URLs). It is now the single source of truth the three import, and it carries the verified live table inline so the next person does not re-derive it.

Checked the neighbouring `/corelink/upgrade?plan=` link rather than assuming the same defect: it answers **307**, not 404. No finding there.

## B — the `undefined` → `null` hunk

The one PKG-3 dropped by contract while flagging it as real. Under `exactOptionalPropertyTypes`, `body: ArrayBuffer | undefined` is not assignable to `RequestInit`; `null` is the spec-correct "no body" for GET/HEAD.

Re-measured against **today's** main — comparing the error **sets**, never the counts:

```
BEFORE 17   AFTER 16
fixed (only before): src/index.ts(2209,59): error TS2379
new (only after):    (none)
```

**A first tsc run reported "0 errors" with exit 1.** `node_modules` was absent and npx had resolved a different package — *"This is not the tsc command you are looking for"*. Caught by the exit-code/content mismatch; the numbers above come from `./node_modules/.bin/tsc`.

## Verification

`vitest` — 2 files, **6 tests, all passing**. Live URLs checked with `curl` (direct status, not following redirects, so a 308→404 chain could not read as success).


gmhelmold:

Gated at `ac4e09d41b49c9ae229a278e67f13c5e2d764b9c`, merged by `scripts/pre-merge-gate-check.sh --merge 1439`.

```
[pass]     build
[pass]     cf:build
[pass]     dco
[pass]     gitleaks detect
[pass]     lint
[pass]     playwright critical-flows (chromium)
[pass]     playwright legacy (chromium)
[pass]     radar
[pass]     test
[pass]     typecheck
[pass]     vitest
```

PR #1440, opened 2026-08-30
fix(observability): an edge-served findMissingBlobs emitted no CAS SLI (B-055)

gmhelmold:
Closes the loop opened by the F2 rollout. **Not mergeable-and-done: the last mile needs a deploy, and I stopped there deliberately** — see *What is NOT proven*.

## The gap

With `EDGE_FIND_MISSING = "on"` in all five prod envs, the Worker answers `findMissingBlobs` in-colo and **never reaches** `R2CasHandler::exists_batch_inner` — which is where `Sli::AvailCasGet` / `Sli::LatencyCasGetP99` are emitted. The audit rows were already written through the container (B-047); the **observation** was not.

So an R2 fault in-colo could have made every edge probe slow and moved those SLOs by **zero**, in every region. That is SLI blindness, not a missing metric: the SLO reports health nobody measured.

## The seam

The emit rides the **awaited** `POST /_internal/audit/cas-attempted` call the edge already makes — the one place the edge and the container agree on what happened. No second transport, no second author.

It mirrors `exists_batch_inner`'s own `emit` closure exactly: **one pair per BATCH, not per digest**, on the same three outcomes — cross-tenant denial and audit failure as errors, a written batch as a success.

Both handlers and the route now fold into **one** process-wide `CountingSliObserver` (`sli_aggregate::shared()`). Two observers would have produced two partial views of a single SLO — the same shape of blindness this item exists to remove.

## The latency number, stated rather than mixed

The Worker reports the **R2 probe window** as `edge_ms`. That window **excludes the audit round-trip being made to report it**, while the container's own latency *includes* its audit write — so edge-served samples run slightly **low** against container-served ones. The alternative is a number that cannot be measured before it is sent.

The field is `Option`, so a Worker deployed before it existed still gets its rows and its `204`; the observation is then recorded with a zero latency, which `SliCounters` keeps distinguishable from "fast" by counting it in `total` while adding nothing to the sum.

## Evidence

| gate | result |
|---|---|
| Rust latency pin, **mutated first** | forcing `latency_us` back to `0` fails with `left: 0, right: 10000` |
| `cargo test -p corelink-server --lib` | **1453 passed / 0 failed** |
| `clippy --all-targets -- -D warnings` | clean |
| `cargo fmt --all -- --check` | clean |
| worker vitest, serve suite | **10/10** |
| `validate_okf` | 1 failure, pre-existing (below) |

Said plainly rather than implied: **removing the denial-path emit still compiles and is not covered by that test** — the test drives `emit_sli` directly, and the branch selection is pinned by the existing status-code tests. It is in the test's own doc comment.

## What is NOT proven — the deploy freeze

That the emit **reaches production**. The container reads its env at process start and this changes container code, so it needs a build + repin + `cf-deploy-prod`, and a freeze is in effect. I did not request one and did not assume one.

The probe that closes it, once a roll is authorised:

```bash
curl -s -o /dev/null -w '%{http_code}\n' -X POST "https://corelink-api.humangr.com/bazel/v2/$TENANT/findMissingBlobs" -H "Authorization: Bearer $PAT" -H 'content-type: application/json' --data-binary @body.json
```

then, on an **unfiltered** `wrangler tail --env prod` (`--search` returns zero bytes with the line present — B-059's sibling defect), look for `edge_find_missing_served`, and in the container log for the `sli aggregate` line carrying `sli=AvailCasGet` with a non-zero `latency_us_sum`. Today that line exists for container-served batches only.

## Pre-existing, not from this branch

`operations.md`'s `limiter.rs` blob anchor fails C4b. Reproduced with this branch's doc changes stashed, so it is inherited from `main`; #1435 is the open fix. The checkpoint bump my reanchor wrote into that file was reverted — it does not belong here.


gmhelmold:

Gated at `edab9ebcc6b2f27d2069ceee5054d58c0f7139d7`, merged by `scripts/pre-merge-gate-check.sh --merge 1440`.

```
[pass]     Cross-validate INV registry × TLA × code × tests
[pass]     PR gate (crate-scoped clippy + tests + canonical vectors)
[pass]     PR gate (crate-scoped clippy + tests)
[skipping] cargo-fuzz nightly (1h)
[pass]     cargo-fuzz smoke (60s)
[pass]     cargo-mutants (changed lines only)
[skipping] cargo-mutants kill-rate ≥ 80%
[pass]     dco
[pass]     gitleaks detect
[pass]     proptest density gate
[pass]     radar
[pass]     vitest
```

PR #1442, opened 2026-08-30
docs(roadmap): the exec-server warning went stale at #521 and nearly blocked a fix

gmhelmold:
`ROADMAP.md` §V.5 states that renaming `EXEC_SERVER_TOKEN` → `EXEC_SERVER_AUTH_TOKEN` *"armaria um `/exec` sem autenticação"*.

**That was true when written. It stopped being true at corelink-runners #521** — *"move the fail-closed gate into the LIBRARY — an open `/exec` is now unconstructible"* — and the document did not follow.

## Verified against the code, not the PR title

```rust
pub fn from_env() -> Result<Self, ExecAuthError> {
    match std::env::var(AUTH_TOKEN_ENV).ok().filter(|t| !t.is_empty()) {
        Some(token) => Self::bearer(token),
        None => Self::unauthenticated_opt_in(),
    }
}
```

- The second arm **errors** unless `CHECK_EXEC_ALLOW_UNAUTH` is set — and nothing in the spawn path sets it.
- `main.rs:22` propagates with `?`, so without a token the process **does not serve**.
- The deployed path is the **binary**: `Dockerfile.runner-devenv:41` builds it, `supervisord.conf:147` runs `/usr/local/bin/exec-server --bind-addr 127.0.0.1:9090` — **loopback**.
- `main.rs` is the only production caller of `app_with_auth`; everything else is tests.

So the rename arms the endpoint **with** authentication. Today it is **dead, not open**.

The paragraph's error was treating the `lib.rs` **API surface** — a caller *could* pass `Unauthenticated` — as though it were the deployed path. The difference between "the function accepts it" and "the binary does it".

## Why the original wording is kept, quoted

Because the correction is the finding. This warning **nearly stopped a correct fix**: I held `corelink-runners#522` rather than push against a written security claim on my own reading, and it took the lead proving the container entrypoint to unblock it. That was the right sequence — but a remediation line that aged without an expiry is indistinguishable from a live one, and false risk spends exactly the same attention as real risk.

The rename itself is HuGR-Labs/corelink-runners#522.

## Gates

`validate_specs` 0 · `validate_okf` 0. `backlog_verify` reports **BROKEN B-061 — verify command exceeded 120s**, which reproduces identically on clean `origin/main` with this change stashed out. Pre-existing, and worth flagging: that is a **third** distinct failure mode for B-061's verify today (drifted-by-count, drifted-by-`validate_okf`-exit, now timeout).


gmhelmold:

Gated at `392388d5a75c5a6e3004046c1803d68dbecb0aec`, merged by `scripts/pre-merge-gate-check.sh --merge 1442`.

```
[pass]     dco
[pass]     gitleaks detect
```

PR #1393, opened 2026-08-27
fix(container): thread the resolved per-tier storage cap into the npm tarball moat (F-008)

gmhelmold:
## F-008 — stale cap after downgrade (npm)

`NpmMoatStore::put` passed a hardcoded `None` storage cap into `MoatCache::put` on every npm tarball write. A tenant writing **only** via npm never re-seeds a lowered cap after a tier downgrade → keeps a stale-high cap. cargo and OCI already avoid this by threading a resolved cap on each write; npm is a clean mirror (it already namespaces per-tenant).

## Fix (mirror the cargo idiom)
- `NpmMoatStore` gains `cap_resolver: Option<Arc<dyn TenantCapResolver>>`.
- `put` resolves the per-tier cap (keyed by the PAT-derived tenant) and threads it into the CAS write instead of `None`.
- `npm::router(...)` takes the resolver as its last param; `routes.rs` injects the same D1-backed `D1TenantCapResolver` cargo/OCI use.

Absence (no resolver, dev/CI) OR an indeterminate cap (D1 fault → `None`) stays `None` — never treated as unlimited (**fail-closed**). Per-tenant namespace unchanged.

## Out of scope
brew/pip store under the `_public` shared namespace, which `byte_accounting.rs:800-808` hard-overrides to `Some(0)` unlimited. Threading a cap there is inert; charging the calling tenant for `_public` writes (**F-007**) is a separate accounting change, tracked separately.

## Verification
- `cargo clippy -p corelink-server --all-targets -- -D warnings` — clean
- `cargo test -p corelink-server --lib npm` — **10 passed, 0 failed**, incl. two new direct-store tests: the resolved cap reaches `CasWriteRequest::storage_quota_bytes`, and `None` when no resolver is wired.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1389, opened 2026-08-27
feat(container): graceful SIGTERM shutdown so a CF rollout drains in-flight requests (F-017)

gmhelmold:
## F-017 — deploy without drain

The container data-plane server did a bare `axum::serve(listener, app).await` (`crates/corelink-container/src/main.rs`). On a Cloudflare containers rollout the platform delivers **SIGTERM**; with no shutdown signal wired, every in-flight HTTP request to the data plane is aborted mid-flight during a roll.

## Fix
`axum::serve(...).with_graceful_shutdown(shutdown_signal())`. `shutdown_signal()` resolves on **SIGTERM** (the rollout signal) or **SIGINT** (local Ctrl-C, dev/test), with a `ctrl_c` fallback when `signal::unix` is unavailable. On the signal the server stops accepting new connections and drains in-flight requests before exit.

## Scope
Drain half only. Tracked separately (TODO(F-017) in-file): the wrangler `containers-rollout` drain policy (default `none`) and a post-deploy `/_health/container` `storage==r2` readiness assertion, plus the ≤120s stale-DO window.

## Verification
- `cargo clippy -p corelink-server --all-targets -- -D warnings` — clean
- `cargo test -p corelink-server --lib` — **1434 passed, 0 failed**

_(A flawed unit test in the drafted change — one that awaited `ctrl_c` without ever raising a signal, so it could never resolve — was removed; a real signal-raise unit test would need a `libc`/`nix` dep and risks the shared self-hosted runner. Signal delivery is validated by the deploy drain, not a unit test.)_

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1460, opened 2026-08-31
docs(backlog): the OKF gate pays for the wrong fix (B-122, B-123)

gmhelmold:
Two items from the #1389 / #1393 citation incident. The first names the defect as **incentive**, not discipline.

## B-122 — a `source_blobs` anchor is vacuously green

| fix | cost | gate result |
|---|---|---|
| advance the anchor | one command | **green immediately** |
| renumber the citations | script + byte-for-byte content verification + manual sweep of the abbreviated citations `CITE_RE` cannot see | green, eventually |

**The root, written down because it was nowhere:** with a `source_blobs` anchor present, C5 compares the tree **with itself** for that file. It is **vacuously true** for the anchored file — it catches drift introduced *after* the anchor and hides the drift the anchor was written *on top of*.

**Evidence, and it is unusually hard for a process item:** on 2026-08-30 the shortcut was taken on **#1389 and #1393, by two independent sessions, both of which knew the caveat** — one of them having written it. Both times `validate_okf` reported **0 stale** while citations pointed at wrong lines: **17 of 18** `main.rs` citations in `planes/container.md`, and **16 more** in #1393. Total recurrence in a sample of two, among people who knew better, is not anecdote — it is a cost ratio.

**The remedy is NOT removing the anchor.** Without it the gate red-flags a file the branch legitimately alters, which is a real need. The anchor simply does not *substitute* for renumbering — two different problems that collapse into one under time pressure.

**#1440's B-059 is the amplifier:** an abbreviated `` `:N-M` `` citation is invisible to `CITE_RE`, so no gate ever reports it. Vacuously-green anchor **plus** blind surface is how 17 wrong citations survive two reviews. Seven of the ones corrected on #1389 were in that form.

## B-123 — hand-edits and a bulk shifter over the same file double-shift it

A bulk shifter computes offsets from the **original** line numbers, so it is not idempotent against a partially hand-corrected file. Reproduced on #1389: `main.rs:48-55` hand-corrected to `49-56`, then the script produced **`50-57`** — a blank line — with the gate green throughout, because the checkpoint had been advanced in the same operation.

Rule is per **file**, not per session: pick one method. If a hand edit already happened, `git checkout -- <file>` back to the base and let the script do all of them.

## What these items do NOT claim

Both `verify` blocks say so explicitly. Neither shows any concept is wrong *today* — only that the cheap wrong fix remains available and rewarded, and that no shared idempotent shifter exists. Documentation only; no gate behaviour changes here.

## Ids

Taken as the next free above the real max on `main` at push time (**121**), not reserved in advance — the `B-NNN` placeholder is not viable because `backlog_verify` rejects a non-numeric id and the item is born `BROKEN`.

`python3 scripts/backlog_verify.py` — **123 items: confirmed=123, drifted=0, stale=0, broken=0**.


gmhelmold:

Gated at `9905133a8cff98274898afa437c5bcba35a83de3`, merged by `scripts/pre-merge-gate-check.sh --merge 1460`.

```
[pass]     dco
[pass]     gitleaks detect
[pass]     verify
```

PR #1461, opened 2026-08-31
docs(audit): D-1 dossier — the audit chain seals 200 rows/hour (B-124)

gmhelmold:
Stacked on #1460 (B-122/B-123) because `backlog_verify` requires **dense** ids — B-124 cannot exist without B-122/B-123 in the tree.

First station of the go-live validation. Measured **against production**, with a control query beside every count, so that "found nothing" is distinguishable from "my instrument broke".

## The finding

| measure | value |
|---|---|
| rows sealed per hour, six consecutive hours | **200, 200, 200, 200, 200, 200** |
| seal latency, last 24 h (n=2692) | min **12 s**, mean **1 h 28 min**, max **5 h 13 min** |
| unsealed backlog | **4 243** of 77 935 |
| oldest unsealed row | **6.8 h** |
| arrivals, same window | 86, 300, 2, 30, 88, 92 (mean ≈ 100/h) |

Six identical integers is a **hard cap**, not a load curve — a load-shaped series does not land on the same integer six times.

The drain currently outpaces mean arrivals, so the backlog shrinks. **The entire safety margin is a factor of two against a bursty process whose observed peak (300/h) already exceeds the ceiling.**

**What the customer gets:** an event is not tamper-evident when it happens. It becomes so **~1.5 h later on average**, and in the observed tail **over 5 h** later. Anything read from `audit_outbox` inside that window is present but outside the chain.

That matters more than it otherwise would because the per-link hash is un-keyed (B-054), so the **head signature carries the whole guarantee** — which makes seal latency exactly the window in which there is none.

## A retraction of my own claim

I previously asserted, **from memory and without measuring**, that `AUDIT_CHAIN_SIGNING_SEED_HEX` sits in plaintext in `wrangler.toml`. **False.**

| query | control | result |
|---|---|---|
| name in `wrangler.toml` | `EDGE_FIND_MISSING` → 5 | **0** |
| name in `corelink-prod` secrets | `CORELINK_AUDIT_ATTEMPTED_AUTH_KEY` → 1, of 53 | **0** |
| `ERASURE_ATTESTATION_SEED_HEX` there | same | **PRESENT** |
| signed chain heads | `audit_outbox` → 77 935 rows | **365 of 365 signed** |

Key handling is **correct**. The dedicated slot is unset and the drain falls back to the erasure-attestation seed by design.

## What this does NOT decide

Stated in the dossier rather than papered over: **where the 200 comes from**. It is not a literal in `audit_drain.rs` — grep returned 0 there against a control of 5 hits for another string in the same file, so the absence is real. The value comes from a caller I did not identify, and guessing it is exactly the kind of claim this campaign keeps retracting. Also undecided: end-to-end offline verification by an outsider, and internal consistency of the sealed segment.

## On B-124's `verify`

Deliberately **structural**, not a threshold. A verify grepping for `200` would go green the moment someone moved the constant without changing the latency; one that queried prod would make the gate depend on the network. The `verify-means` says so explicitly, and points at the dossier's queries as the real check.

`python3 scripts/backlog_verify.py` — **124 items: confirmed=124, drifted=0, stale=0, broken=0**.


gmhelmold:

Gated at `345fa9407a399125bf36907fd5fe98843c2ec253`, merged by `scripts/pre-merge-gate-check.sh --merge 1461`.

```
[pass]     dco
[pass]     gitleaks detect
[pass]     verify
```

PR #1465, opened 2026-08-31
docs(audit): D-2 dossier — residency is UNEVALUABLE for 3 670 rows (B-126)

gmhelmold:
Third in the stack: #1460 → #1461 → this. Stacked because `backlog_verify` requires **dense** ids — B-126 cannot exist without B-123..B-125 in the tree. **Merge order is mandatory, not a preference.**

**Version stamp:** measured against production D1 while the five prod containers ran **`cab16a3a-r1`** — **21 commits behind `main`, 5 of them container-affecting**. These numbers describe the code that is *serving*, not the code that is merged.

## The finding is not the violation count

| query | result |
|---|---|
| audit rows whose `region` ≠ tenant's `primary_region` | **0 violations** (over 74 333 joined rows) |
| rows in `audit_outbox` | **77 935** |
| **rows whose `tenant_id` has NO `tenant` row** | **3 670** — 175 distinct ids, **4.7%** |

The clean number was computed over a population **the query itself had already trimmed**. A `JOIN`-based residency check cannot evaluate its predicate for a row with no tenant — and it drops those rows *without saying so*.

**Unevaluable is a third state**: weaker than satisfied, different from violated, and nothing in the system currently tells it apart from satisfied.

## Data in a region no tenant is provisioned for

| | |
|---|---|
| tenants by `primary_region` | enam **153**, wnam **108**, apac **1** — **none in `weur`** |
| audit rows by `region` | enam 76 129, wnam 1 869, **weur 4**, apac 4 |
| do the 4 `weur` rows have a tenant? | **no — 0 of 4** |

Trivial count, non-trivial property: a row exists in a region whose residency predicate has nothing to check against, and the JOIN check never saw it.

## What bounds any green result here

**76 129 of 77 935 rows (97.7%) are `enam`.** The cross-region rejection path — 409 `residency_violation`, plus migration-0023's `RAISE(ABORT)` triggers — is real code that production traffic has barely exercised. **A control that never fires has not been shown to fire.** The 0 is evidence that nothing went wrong, not evidence that the mechanism would catch it.

## What this does NOT decide

Listed rather than glossed: **object-plane residency** (the CAS tenant prefix is an HMAC under a write-only TDK, so a listing cannot be attributed to a tenant); **whether the 409 path fires** (needs a deliberate cross-region request against prod); **the two distinct `Region` enums** (no finding here depends on telling them apart — every query reads the `region` *column*); and **R2 Object Lock**, which returns `NotImplemented` and is platform-blocked rather than unbuilt.

## On B-126's `verify`

Structural, for the same two reasons as B-125's: a hard-coded `3670` would go green on any change to the number rather than on the defect being fixed, and a prod query would make the gate depend on the network and on credentials. The `verify-means` says so and points at the dossier's queries as the real check.

`python3 scripts/backlog_verify.py` — **126 items: confirmed=126, drifted=0, stale=0, broken=0**. Ids dense 1..126, no duplicates, verified **after** the commit.

Docs only — no `crates/`, `Dockerfile`, `Cargo.lock` or `Cargo.toml`, so it does not touch the container pin the deploy is waiting on.


gmhelmold:

Gated at `88b97c207ce7d01e8211266dce56f93f6b04c104`, merged by `scripts/pre-merge-gate-check.sh --merge 1465`.

```
[pass]     dco
[pass]     gitleaks detect
[pass]     verify
```

PR #1496, opened 2026-08-31
chore(okf): resolve abbreviated citations, and record that 43% of them land nowhere (id não alocado)

gmhelmold:
> ⛔ **BLOQUEADO POR ALOCAÇÃO DE ID.** O resolvedor (`scripts/okf_resolve_abbrev_cites.py`) está pronto e commitado; **o item do `BACKLOG.md` não entra neste commit** — hoje não existe id válido para ele. Corrente de backlog aberta: **#1504 (B-131..135) → #1509 (B-136/137) → #1511 (B-138) → #1512 (B-139..141)**. Este PR nasceu pedindo `B-130` (já ocupado na `main`); a realocação para `B-140` **caducou** antes do push.

## What this is

[B-059] records that `CITE_RE` in `scripts/validate_okf.py` requires a non-empty
path, so an **abbreviated citation** — a bare `` `:167` `` written to avoid
repeating a long path in the same sentence — is invisible to C5, C6 and C6b.

That item reads as a counting gap. Measured, it is not one.

## The measurement

`scripts/okf_resolve_abbrev_cites.py` (added here) resolves each abbreviated
citation the way a reader does — against the nearest preceding full-path citation
in the same line — and reports where it actually lands. Run against the whole
wiki on `origin/main`:

| failure mode | count |
|---|---|
| lands on a comment | 20 |
| lands on a bare `}` / `)` | 12 |
| lands **past the end** of the inherited file | 6 |
| lands on a blank line | 5 |
| inherits no path at all | 3 |
| **total unresolvable** | **46 of 107, across 12 concepts** |

**43% of the abbreviated form points at nothing, and no gate sees it.**

## Why this is not about counting

The six past-EOF ones are the argument. `crates/handler-trait-seam.md:60` names
`customer_d1.rs` in prose, **without a line number**, then writes `` `:947` ``,
`` `:1011` ``, `` `:1086` ``, `` `:1203` ``, `` `:1428` ``, `` `:1646` ``.

A human reads those as `customer_d1.rs` and is right. The nearest full-path
citation is `corelink-handler-customer/src/handler.rs`, which has **804 lines** —
so every one of the six points past the end of the file it mechanically inherits.
Readable by a person, meaningless to a machine, and nothing says a word.

The other 40 are already forbidden by the frozen contract, which requires a cite
range to contain a real CODE line, "non-blank, not a `//`/`//!`/`///`/`#`/`*`/`/*`
comment". That rule is enforced for full-path citations and silently skipped for
abbreviated ones. **The abbreviated form is not shorthand — it is an exemption
from the rule the wiki is gated on.**

## The concrete regression that motivated it

#1472 moved `build_with_factory` out of `routes.rs` and orphaned five abbreviated
citations, one of them carrying a behavioural invariant: byte accounting wraps
*inside* the erasure gate, so a tombstoned re-PUT is refused before bytes are
reserved. Reversing that order would reserve bytes for a write the gate then
rejects.

`validate_okf.py` reported the profile **valid before the breakage and valid
after it**. The regression was found by reading, not by the gate.

## The gate discipline this script follows

**Anti-vacuous.** Zero concepts found, an unreadable concept, or zero abbreviated
citations in a full scan each exit **FATAL (2)**, not green — the wiki uses the
form freely, so finding none means the scan broke, not that the wiki is clean.

**Named failure modes**, not a bare count: `no-inherited-path`, `file-missing`,
`past-eof`, `blank`, `comment`, `delimiter`.

**Calibrated before its output was believed.** A citation that resolves to real
code is not flagged (positive control); rewriting that same citation to a line
past EOF is flagged at both of its occurrences, and restoring leaves a zero-line
diff (teeth test).

**The judgement call is stated, not buried.** The resolver inherits a path only
within the same LINE, while a reader plausibly carries it further down a bulleted
list. Only the 3 `no-inherited-path` findings depend on that choice; the other 43
hold under either reading.

## Scope

Adds the script and the item. **Fixes none of the 46** — they are inherited, and
a PR that fixes what it broke is reviewable while one that also fixes 46 unrelated
things is not.

## The id

O id ainda **não** foi alocado (ver o topo deste PR). O
`verify` is the resolver itself, in `open` polarity — it passes while at least one
abbreviated citation still fails to resolve, and inverts to drop the `!` on close.

## Como alocar o id, quando desbloquear

O id sai **disto**, no momento do push, e **reconfere-se depois de qualquer rebase** — nunca de um número entregue por antecipação:

```bash
git fetch origin main
git show origin/main:BACKLOG.md | grep -oE '^### B-[0-9]+' | sort -V | tail -1
```

**Não empilhe neste PR.** A corrente já tem quatro níveis; um quinto transforma a ordem de merge num campo minado.

**Por que não ficou um placeholder no `BACKLOG.md`.** Testei: `id: B-UNALLOCATED` sai **CONFIRMED** e a checagem de densidade nem o vê — ela só olha `B-\\d+`. O placeholder mergearia em silêncio, que é a classe de defeito dominante deste repo. Falhar por ausência é melhor que passar por omissão, então o bloco fica **fora** do commit e vai abaixo, pronto para colar.

## O item, pronto para colar (só falta o número)

<details><summary>bloco do <code>BACKLOG.md</code></summary>

````markdown
### <ID-A-ALOCAR> — an abbreviated citation inherits a path the prose contradicts, so 43% of them point at nothing

[B-059] records that `CITE_RE` in `scripts/validate_okf.py` requires a non-empty
path, so an abbreviated citation — a bare `` `:167` `` written to avoid repeating
a long path in the same sentence — is invisible to C5, C6 and C6b. That item is
correct but understates the failure. It reads as a counting gap. It is not.

Measured on `origin/main` with `scripts/okf_resolve_abbrev_cites.py`:
**46 of 107 abbreviated citations across 12 concepts do not resolve to a line of
code at all.** By failure mode: 20 land on a comment, 12 on a bare `}` or `)`,
6 past the end of the file, 5 on a blank line, and 3 inherit no path whatsoever.

The `past-eof` six are the ones that show why this is not a counting gap.
`crates/handler-trait-seam.md:60` names `customer_d1.rs` in prose, WITHOUT a line
number, then writes `` `:947` ``, `` `:1011` ``, `` `:1086` ``, `` `:1203` ``,
`` `:1428` ``, `` `:1646` ``. A human reads those as `customer_d1.rs` and is
right. The nearest full-path citation is `corelink-handler-customer/src/handler.rs`,
which has 804 lines — so every one of the six points past the end of the file it
mechanically inherits. The citation is simultaneously readable by a person and
meaningless to a machine, and no gate says a word.

The contract already forbids the other 40: §4 requires a cite range to contain a
real CODE line, "non-blank, not a `//`/`//!`/`///`/`#`/`*`/`/*` comment". That
rule is enforced for full-path citations and silently skipped for abbreviated
ones, so the abbreviated form is not shorthand — it is an exemption from the rule
the wiki is gated on.

This is the house's signature shape, in its purest form: **the gate is green, and
was green the whole time.** A router-move PR (#1472) orphaned five abbreviated
citations, one of them carrying a behavioural invariant — that byte accounting
wraps INSIDE the erasure gate, so a tombstoned re-PUT is refused before bytes are
reserved. `validate_okf.py` reported the profile valid before the breakage and
after it. The regression was found by reading, not by the gate.

`scripts/okf_resolve_abbrev_cites.py` (added with this item) is the missing half.
It resolves each abbreviated citation against the nearest preceding full-path one
in the same line and classifies where it lands, naming each failure mode rather
than returning a bare count. It is anti-vacuous: zero concepts, an unreadable
concept, or zero abbreviated citations found in a full scan are each a FATAL
scan-failure exit, because the wiki uses the form freely and finding none would
mean the scan broke, not that the wiki is clean. Calibrated before use — a
citation resolving to real code is not flagged, and rewriting that same citation
to a line past EOF is flagged at both of its occurrences.

Closing this means every abbreviated citation resolves to a real code line, and
the resolver runs where a broken one can block a merge. Note the one judgement
call the resolver makes: it inherits a path only within the same LINE. A reader
plausibly carries the path further down a bulleted list. Only the 3
`no-inherited-path` findings depend on that choice; the other 43 hold under
either reading.

```backlog
id: <ID-A-ALOCAR>
repo: corelink-server
owner: tl
status: open
verify: |
  ! python3 scripts/okf_resolve_abbrev_cites.py --quiet
verify-means: |
  open — the resolver exits non-zero, i.e. at least one abbreviated citation in
  `docs/knowledge/` still resolves to a comment, a delimiter, a blank line, a
  position past the end of its inherited file, or to no file at all. Closes when
  the resolver exits 0, at which point this `verify` inverts to drop the `!`.
  A FATAL exit (2) is a broken scan, not a pass, and fails this check too.
last-verified: 2026-08-31
```
````

</details>


gmhelmold:
Closing as superseded by #1575 (merge eee6e77d, landed B-059 head ce441206). The abbreviated-citation resolver was incorporated into main and then hardened with bounded parsing, path/context binding, range integrity, fail-closed workflow wiring, mutation coverage, and backlog reconciliation. This PR changes only the earlier resolver file and has no unique implementation left to preserve. Do not cherry-pick.

PR #1508, opened 2026-08-31
test(container): cover the pip F27 write gate and its REV-S3 attribution

gmhelmold:
> **Stacked on #1486** (`claude/wp-t1-pip`). `routes/pip/` does not exist on `main`, so this
> targets the split branch. Rebase onto `main` once #1486 lands.

## The gap

`pip_gate`'s write arm had **no test at all**. Positive control:

```
grep -rn "Method::PUT\|resolve_with_capability\|can_write" crates/corelink-container/src/routes/pip.rs   -> 11 hits
grep -rn "Method::PUT\|resolve_with_capability\|can_write" crates/corelink-container/src/routes/pip/     ->  0 hits
```

Every existing pip test is a GET or an unmapped method, so `PipPatResolver::resolve_with_capability`
was **never called once**. That is ~33 lines of fail-closed security code plus the authoritative
cost-attribution capture, and nothing would have gone red if any of it regressed.

Found while splitting `pip.rs` for B-126 (#1486). That PR deliberately only MOVED tests, so this is
separate work.

## What was added

Six tests in a new `tests_gate_write.rs` — its own file because it is **one property**: *a write is
admitted only when the PAT's own D1 row authorises it, and is billed to the tenant that row names.*

| test | what it pins |
|---|---|
| `put_without_bearer_token_is_401` | the scope header alone is never enough to write |
| `put_with_read_only_pat_is_403_despite_rw_scope_header` | **the F27 case**: header says rw, D1 row says no — the row wins |
| `put_with_unverifiable_pat_is_401_fail_closed` | a resolver error is a denial, not a pass-through |
| `put_with_write_capable_pat_clears_the_gate` | both layers agree, the gate hands the request on (405 — pip has no PUT route) |
| `write_is_billed_to_the_pat_tenant_not_the_forged_header` | **REV-S3**, read back from the quota store |
| `write_has_an_attribution_key_without_any_tenant_header` | the differential that proves where the key comes from |

## Three things about the surface that shaped them

**Every test sends `x-corelink-scope: cas:rw`.** A PUT without it is refused by the *scope* branch one
step earlier, with its own 403, and would prove nothing about F27. Holding the header permissive is
what leaves the PAT's D1 row as the only thing deciding — which is precisely the threat model F27
exists for: a Worker that stamps a write scope onto a PAT whose record does not grant one.

**The `can_write` bit is `requires_cache_write(&row.scope)`** (`adapter_pat.rs:1986`), so the read-only
PAT is minted `SCOPE_CACHE_RW` and differs **only** in its D1 row scope (`cas:r`). That mirrors
`adapter_pat`'s own `memo_caches_no_authorization_state` and models the real case — a token already in
a customer's hands, narrowed in the database afterwards.

**REV-S3 cannot be proven from a status code.** The fixture charges \$1/op against a fresh tenant,
admitted under either label, so an attribution bug is invisible in the response. Added
`router_with_quota_observable`, which returns the quota store so the test can assert **who** was
charged; `router_with_quota` now delegates to it and the existing test is untouched. Confirmed first
that the guard **seeds a row for a brand-new tenant** — otherwise an empty store would have meant a
wrong assertion rather than a wrong system.

The last test is the differential. The sibling property file pins that a READ with no tenant header
fails CLOSED with 503, because a read runs no PAT verification in this gate and has no other source
for the label. A WRITE with no tenant header must **not** 503. Same gate, same missing header,
opposite outcome — and the only difference is the F27 verify.

## Proof the tests bite

Mutation-probed, restored between runs:

| mutation | result |
|---|---|
| drop the `Ok(_no_write)` denial (fall through) | **12 passed, 1 failed** — `put_with_read_only_pat_is_403_despite_rw_scope_header` |
| drop the REV-S3 tenant capture | **11 passed, 2 failed** — both attribution tests |

Each mutation reddens **exactly** the tests that claim that behaviour, and no others.

## Gates

- `cargo test -p corelink-server --lib routes::pip` — **13 passed, 0 failed**
- `cargo clippy -p corelink-server --all-targets -- -D warnings` — exit 0
- `cargo fmt --all --check` — clean
- **No production behaviour changed.** `pip.rs` gains only a `mod` declaration and the comment listing
  the property files. The 8 distinct OKF ranges citing `pip.rs` are byte-identical, with a negative
  control that flags all 8 on a one-line shift.

Generated with [Claude Code](https://claude.com/claude-code)


gmhelmold:
Closing as superseded by merged #1576 (`0f5cb28af`, exact landed head `1f7dd4abe`). The six original F27/REV-S3 cases remain in the landed suite; the valid-write control was strengthened to an observable downstream 204 probe, resolver backend/load-shed behavior gained an explicit fail-closed 503 + Retry-After case, attribution remains store-observable, and current OKF citations were reconciled. No unique behavior from this stale-base head remains. Do not cherry-pick.

PR #1540, opened 2026-08-31
docs(skills): three maintenance skills for the defect classes that cost the most time

gmhelmold:
Three skills under `.claude/skills/`, written from the measured incidents of this campaign rather than from principle. Each is a class we have paid for more than once.

## `verify-population`

The dominant defect class: **the number is correct about the set it measures, and that set is not what the sentence promises.** Six measured instances are tabled in the skill — downloaded vs installed, run-level vs job-level, mentions vs uses, files-touched vs surface, scheduled checks vs checks that should exist, and compiled-in vs serves-this-tenant.

It carries the reason no gate catches it: tests check the number is computed correctly, review checks the sentence is well written, **nobody checks the join**. And three detectors ordered by how much domain knowledge they demand, so the first one works with none — a degenerate output (zero variance, 100% pass, `X-Y` always 0) accuses the instrument before the world.

## `teeth-test`

A green gate proves the code is correct **or** that the gate cannot fail, and does not say which.

Includes the false pass measured this week: a `verify` replacement whose text-splicing spliced into the wrong block, deleted a neighbouring backlog item, and still exited 0 with CONFIRMED. Caught only by comparing the printed string, not the exit code. Also the uncovered-comparator hole (a selftest that stays green at N/N when its own comparison is replaced with `if true`), the discriminating DoD for a test file, anti-vacuity clauses for a counting gate, and the seven instruments in this toolbox that report absence when the answer is something else.

## `built-not-wired`

Existing is not shipping. Three states get reported with the same sentence, and reporting "built, not wired" as "not built" throws away work that already exists and points the plan at the wrong task.

Reachability is a property of the **target**: this repo has 19 target-gated dependency blocks, so textual `Cargo.toml` parsing invents edges no build has, and `cargo tree` suggests `--target all` — which returns paths that exist in no build — exactly when the correct answer looks empty.

It also carries the label-versus-dispatch misread from this week: a `const fn` named for the active provider whose only call sites were `/healthz`, a metrics label and a log field was read as the dispatch, producing the wrong conclusion that the architecture was single-provider. The dispatch was already a trait object, and a sibling module already held several providers at once.

## Notes

- Docs only. No code, no workflow, no gate touched.
- Follows the existing `.claude/skills/<name>/SKILL.md` convention (`okf-context`, `okf-reconcile`, `techlead`), with trigger-rich descriptions so they load when relevant rather than on request.
- The three cross-link each other; they share a root cause.


gmhelmold:

Gated at `b0e7bdf63ad5dabb8008e59a3c7d98cc6143bf4c`, merged by `scripts/pre-merge-gate-check.sh --merge 1540`.

```
[pass]     dco
[pass]     gitleaks detect
```

PR #1533, opened 2026-08-31
fix(docs): 9 links mortos nas páginas legais vivas — e o portão que os deixou passar (B-168)

gmhelmold:
## The finding

`apps/docs/docusaurus.config.ts:148` sets `onBrokenLinks: "throw"` and the docs
build is green. That setting only covers links Docusaurus itself resolves —
Markdown/MDX links and `<Link>`. A raw `<a href="/...">` typed into a React page
under `apps/docs/src/pages/**` is invisible to it. The `broken-links` job
(lychee, against the build output) did not catch these either.

Confirmed against **production**, not by reading:

```
$ curl -s -o /dev/null -w '%{http_code}\n' https://humangr.com/corelink/docs/explanation/compliance/dpa
404
$ curl -s -o /dev/null -w '%{http_code}\n' https://humangr.com/corelink/docs/explanation/sre/slo
404
$ curl -s -o /dev/null -w '%{http_code}\n' https://humangr.com/corelink/docs/legal/terms
200
```

The brief named 4 dead links. A full sweep found **9**, and two of them are in
`legal/privacy.tsx`, not `trust.tsx` as reported (`trust.tsx` has exactly one
`href`, a `mailto:`).

## What changed, per link — before → after

| # | file:line (pre-fix) | before | after | why |
| --- | --- | --- | --- | --- |
| 1 | `legal/terms.tsx:186` (§5) | `<a href="/explanation/compliance/dpa">Data Processing Addendum</a>` | plain text `Data Processing Addendum` + "The DPA is not published on this site; request a counter-signed copy at legal@humangr.com" | no published destination |
| 2 | `legal/terms.tsx:311` (§13) | same href | plain text `Data Processing Addendum` (it is a named contract in an entire-agreement clause, not a page) | no published destination |
| 3 | `legal/terms.tsx:332` (**§14**) | `documented at <a href="/explanation/sre/slo">SLO targets</a> and reported publicly on the Statuspage.` | `reported publicly on the Statuspage.` | `docs/explanation/sre/` does not exist and **no published page documents the probe set** — the deleted clause was the only untrue part |
| 4 | `legal/terms.tsx:391` (footnote) | same href | `<a href="mailto:legal@humangr.com">Data Processing Addendum (on request)</a>` | keeps the affordance, drops the false page |
[...]

PR #1544, opened 2026-08-31
fix(gitleaks): match the SHAPE of a hex secret, not a list of variable names

gmhelmold:
## The hole, measured

`.gitleaks.toml` carried `corelink-internal-auth-hex`, whose regex **enumerated exactly two
name patterns** — `CORELINK_[A-Z_]*AUTH_KEY` and `PAT_SIGNING_KEY[A-Z_]*`. Every other
`openssl rand -hex 32` credential this codebase mints was invisible to it. Not entropy, not the
value, not the path: **the name simply was not on the list.**

Verified before fixing anything, with the *same* version-pinned gitleaks 8.30.1 that
`.github/workflows/gitleaks.yml` installs. A `[env.prod]` `vars` block assigning a 64-hex
literal to `AUDIT_CHAIN_SIGNING_SEED_HEX` returns **zero findings on 3/3 high-entropy trials**,
while the same scan with the same config fires correctly on `CORELINK_INTERNAL_AUTH_KEY`. The
instrument was working; the criterion was not.

## ⚠️ The premise was understated — the seed is already on `main`

The brief said the seed exists only on PR #1397's branch. **It does not.** The identical
64-hex value (compared by hash of the value, never transcribed) is committed to `main` in
`reports/audits/2026-08-25-comprehensive-audit-and-verification.md:103`, landed by
**PR #1415** (merge `fc3c6e4a`, 2026-08-30) — *an audit report that quoted verbatim the signing
seed it was reporting.*

That PR **ran `gitleaks detect` and it passed in 8 s.** The scanner ran, was green, and was
wrong. This rule is the sole thing that stood between that value and `main`.

## The fix: a criterion, not a list

`corelink-secret-shaped-hex` — a **32- or 64-hex literal assigned to an identifier whose name
contains** SEED, SIGNING, SECRET, PRIVATE, AUTH_KEY, API_KEY or TOKEN. Both previously-matched
names are a **subset** of that set, so this is a proved widening and not a swap.

| scan | old config | new config |
|---|---|---|
| worktree at HEAD (69.8 MB) | 0 | **1** |
| full history of `origin/main` (3,438 commits) | 0 | **1** |

The single new finding **is** the true positive above. **Zero findings lost, zero false
positives.** The discriminator is the NAME, which is what makes matching 64-hex safe at all:
git SHAs (40-hex and bare), `GITLEAKS_SHA256:`, `Cargo.lock` `checksum =`, `sha256:`/`blake3`
digests and lockfile integrity hashes carry none of those words, so they never reach the rule.

**No path was allowlisted for this rule** — none was needed, because nothing false fired. It
inherits the repo-wide fixture allowlist (`tests?/`, `specs/`, `apps/docs/`, `conformance/`,
`openapi/`, `tools/cli/tests/`) unchanged; operational trees stay in scope. The only allowlist
added is value-shaped: degenerate placeholder hex (all-zeros/ones/`deadbeef` repeats), which a
real `openssl rand` value cannot take.

## Teeth test — mutated BY CONTENT, each mutation confirmed before running

Every mutation asserts its own application **inside the rule block** before the check runs.
One earlier attempt silently failed to apply (shell expansion) and reported a **false pass**;
it was redone in Python with a content assertion — which is exactly why this discipline exists.

| mutation | result |
|---|---|
| baseline | `exit 0` |
| revert regex to the old enumeration | `exit 2` — 4 positives stop matching |
| drop only `seed` from the criterion | `exit 2` |
| corrupt the regex (uncompilable) | `exit 2` — broken instrument |
| rename the rule `id` | `exit 2` — rule gone |
| redact the seed from the file | `exit 1` — close and invert polarity |
| restore byte-for-byte | `exit 0` |

Scanner-level cases, run through real gitleaks: the `AUDIT_CHAIN_SIGNING_SEED_HEX` line
**0 → 1**; `CORELINK_*AUTH_KEY` + `PAT_SIGNING_KEY*` regression control **2 → 2**; negative
controls (bare SHA, `sha256:`, `blake3`, `checksum`, `GITLEAKS_SHA256`, placeholder hex)
**0 → 0**.

## Backlog

- **B-168** — the leaked seed is on `main` and **un-rotated**. A `git rm` cannot fix this:
  `gitleaks detect` scans the commit graph and `fc3c6e4a` is immutable. Closing it means
  rotating the secret and bumping `AUDIT_CHAIN_SIGNING_KEY_ID`, then pinning a
  revoked-credential closure in the same rigorous format as the existing `whsec_` block.
  ⚠️ Rotation without a two-`key_id` window breaks verification of the 193 signed partition heads.
- **B-169** — a CONFLICTING PR is never content-scanned (#1397 shows **5** metadata checks
  against a healthy sibling's **31**; `gitleaks` is not among them, and an absent check never
  becomes pending). **Measured and deliberately downgraded:** `pre-merge-gate-check.sh` already
  has "Defense 2" (`REQUIRED_PRESENT = ["dco", "gitleaks", "changelog"]`, STRUCTURAL and
  non-overridable), and it correctly refuses #1397. So the merge-time mechanism one would
  propose **already exists**; the residual is observability only, and this was **not** the
  ingress path — the scanner rule was. A `pull_request_target` scanning lane is described in
  the item and deliberately **not built here**: it runs with base-repo credentials over
  third-party code and is a design decision, not a regex fix.

Both verify predicates were themselves teeth-tested by mutation; both report `CONFIRMED` under
`scripts/backlog_verify.py --id`.

> ℹ️ **Expect the weekly full-history sweep to go red** on the one true positive until B-168 is
> rotated and closed. That red is the finding, not a regression. The PR lane
> (`base..head`) is green on this branch.

## Gates

- `python3 scripts/validate_specs.py` → `475 + 11 = 486, 0 failures`
- `scripts/backlog_verify.py --id B-168` / `--id B-169` → `CONFIRMED` (used `--id`, so a single
  `verify` runs — not the external-tool matrix)
- gitleaks PR lane on this branch → **no leaks found**
- No workflow touched; no `runs-on` changed.
- **The seed value appears nowhere** in this PR — not in the config, the commit message, either
  backlog item, or this body. Verified: 0 hex literals ≥32 chars in the diff.

## Ids

B-168/B-169 were allocated at push time against `origin/main` (max `B-167`). ⚠️ Open PRs
**#1533** and **#1542** also claim `B-168`/`B-169`. Id collisions do **not** produce a merge
conflict, and `backlog_verify.py` requires ids to be **dense**, so 170/171 is not a legal
workaround. Whichever of the three lands second must renumber on rebase.


gmhelmold:
Superseded by merged D03 closure PR #1587 at 953e052af. The final tree contains the secret-shaped-hex gitleaks detector plus history/regression teeth-tests. This closes the duplicate implementation PR; it does not assert unrelated credential-rotation work.

PR #1554, opened 2026-09-01
feat(audit): gate B-101 finding coverage

gmhelmold:
## Summary

- ingest all 110 live audit findings from the three pinned sources
- map 32 findings to canonical backlog IDs and explicitly decline 78 with reason and price
- add deterministic parser, manifest, negative tests and a self-hosted workflow triggered by source, manifest and BACKLOG drift

## Verification

- `pytest -q tests/test_audit_finding_coverage.py` (6 passed)
- `python3 scripts/verify_audit_finding_coverage.py --format json` (110 complete)
- actionlint, gitleaks range, diff-check and DCO
- independent cold review: LAND

B-101 remains open until the new instrument lands and its backlog polarity is closed separately.


gmhelmold:
Cold integrity disposition: REJECT/HOLD; this head must not land. The gate accepts remapping a finding to any existing canonical ID, accepts non-semantic declined metadata such as reason/price `x`, and can be satisfied by rewording the audited source plus updating its digest. Its workflow trigger also omits a required source path. The recovery-integrity stack deliberately excludes this false-green gate and keeps B-101 open/manual until a semantically bound verifier exists. Branch preserved for provenance.

Issue #1633, opened 2026-09-17
[Audit][P2] Production CAS accounting omits the BYOK configuration used by the encrypted storage handler


gmhelmold:
## Finding / evidence level
**Static construction-path proof**, not a live KMS or Rust execution. Applies when the real-provider BYOK path is enabled for an active tenant; no claim is made about how many production tenants currently have BYOK active.

Baseline `cca798ff5bc2df660ecf2570ed243eb9775ff3d0`, inspected 2026-09-17.

## Source chain
1. [R2 CAS factory](https://github.com/HuGR-Labs/corelink-server/blob/cca798ff5bc2df660ecf2570ed243eb9775ff3d0/crates/corelink-container/src/storage/r2_s3_parts/cas_builder.rs#L105-L167) constructs the real KMS/config/TCS/envelope collaborators in a BYOK-enabled image and returns `handler.with_byok(config,tcs).with_byok_random(mode_b)`.
2. [Router composition](https://github.com/HuGR-Labs/corelink-server/blob/cca798ff5bc2df660ecf2570ed243eb9775ff3d0/crates/corelink-container/src/routes/build.rs#L96-L136) wraps this raw handler in `AccountingCasHandler::new(...)` but does **not** call the decorator's `with_byok(...)` or pass equivalent pinned crypto context.
3. [Decorator constructor and helper](https://github.com/HuGR-Labs/corelink-server/blob/cca798ff5bc2df660ecf2570ed243eb9775ff3d0/crates/corelink-container/src/byte_accounting/b126_m2_impl_01_part_02.rs#L34-L98) leave `byok_config_cache=None`; `byok_committed_len(None,...)` returns plaintext length. The method that attaches a config exists but is not used by this composition.
4. [Actual accounting write](https://github.com/HuGR-Labs/corelink-server/blob/cca798ff5bc2df660ecf2570ed243eb9775ff3d0/crates/corelink-container/src/byte_accounting/b126_m2_impl_01_part_02.rs#L232-L337) reserves this amount.
5. [Actual CAS delete](https://github.com/HuGR-Labs/corelink-server/blob/cca798ff5bc2df660ecf2570ed243eb9775ff3d0/crates/corelink-container/src/storage/r2_s3_parts/ac_core.rs) obtains reclaimed size from the physical R2 object, and the decorator releases that returned size.

## Counterexample
[...]

PR #1800, opened 2026-09-22
fix(storage): pin BYOK accounting operation

gustavomhss:
Refs #1633

Pins one BYOK write operation before physical-byte reservation. The pin owns the tenant-scoped data intent and its authoritative config/CMK-version snapshot; accounting derives the stored-byte overhead from that snapshot and the R2 CAS/AC handler consumes the same guard for encryption, publication, and release.

This removes the cache-only handoff that allowed a rotation between reservation and the inner `ByokDataGuard` acquisition. Pins are single-use, tenant-bound, and dropped/released on reservation rejection, rollback, errors, and successful completion. Public content remains plaintext.

The focused hosted gate covers pinned rotation/no-reuse, Mode A and Mode B physical sizes, deduplicated retry rollback, delete reconciliation, public namespace, factory composition, and ambiguous AC liability retention.

Validation is hosted only; no local build, lint, or test was run.


gustavomhss:
Hosted run 35684693854 (focused-accounting) terminated **FAIL** (exit 101) before executing the six exact regressions: compilation failed in `crates/corelink-container/src/storage/r2_s3_parts/tests_3.rs:318` because `BYOK_CLB1_OVERHEAD` was not imported; rustc suggested `use crate::storage::byok_cas::BYOK_CLB1_OVERHEAD;`.

Separately, the cold architectural review still leaves a P1 blocker: operation-pinned BYOK reservation/storage accounting is not modeled end to end. The reservation decision must carry the operation's pinned storage/config identity through the write/delete lifecycle and persist/reconcile against that same identity, including failure and retry paths. Please create a child issue for that contract (operation-pinned reservation identity, durable storage metadata, and retry/reconciliation tests) and link it here before merge consideration. This PR remains unmerged pending that work and lead review.

PR #1825, opened 2026-09-22
test(observability): pin B-107 storage phase separation

gustavomhss:
## Summary

Pins the B-107 storage attribution seam in the existing `MoatCache::put` proof.
The hermetic test gives the R2/CAS window and the URL-map accounting window separate bounded delays, then asserts that both phases are emitted and that their sum stays within the request wall clock.

This preserves the attribution boundary already present on `main`:

- `ostore` is the R2/CAS storage window.
- `oaccounting` is the D1 accounting and URL-map window.
- same-phase re-entry remains recorded once across `spawn_blocking`.

The test does not claim that the historical 654 ms aggregate has improved. Issue #1663 remains open for the authenticated production p50/p90/p99 readback with a pinned deployed version.

## Hosted test manifest (credentialless)

Run on a GitHub-hosted `ubuntu-24.04` runner at this PR head:

```sh
set -euo pipefail

# Use the repository-pinned Rust toolchain and a pinned protoc action in the
# hosted workflow before this command.
rustup show active-toolchain

test_name='adapter_cache::tests::put_records_the_store_phase_exactly_once'
cargo test --locked -p corelink-server --lib "$test_name" --exact --nocapture
```

Expected result:

```text
test adapter_cache::tests::put_records_the_store_phase_exactly_once ... ok
```

Safety bounds: no PAT, D1, R2, Worker, deployment, load run, campaign file, or external endpoint is used. The test uses only in-memory handlers and bounded 30 ms delays.

## Scope and DoD

- One focused test file: `crates/corelink-container/src/adapter_cache.rs`.
- 47 added lines, 48 changed lines total; well below the 400 LOC limit.
- DCO trailer present on the commit.
- Refs #1663 present in the commit subject.
- No campaign edit or production mutation.
- Production closure remains gated on three authenticated warm 1 KiB PUTs with p50/p90/p99 and a recorded deployed version.


gmhelmold:
Retriggering required hosted contexts before the authorized product merge.

PR #1833, opened 2026-09-22
feat(observability): add B-129 diagnostic probe

gustavomhss:
## Summary

- Adds a manually dispatched, bounded B-129 diagnostic lane for #1671.
- Performs up to 10 authenticated HTTPS GETs and emits a redacted receipt with wire commit identity, timing reconciliation, region/request identifiers, and residual bound.
- Keeps the lane read-only: no deploy, write, schedule, or issue-closing action.

## Linked work

- Refs: #1671
- B-129 remains open until the deployed diagnostic flag, dogfood secret, and fresh production evidence satisfy the canonical closure contract.

## Test plan

- `python3 -m py_compile scripts/probe_i1671_b129.py` — passed.
- `git diff --check` — passed.
- No workflow dispatch performed; no local CI run (manual diagnostic only).

## Risk / rollback

- The workflow is inert until explicitly dispatched with owner approval, a deployed SHA, a dogfood tenant/key, and the protected `CORELINK_DOGFOOD_PAT` secret.
- Rollback is deleting the workflow and probe files or reverting this commit.


gustavomhss (commented):
Review note for B-129: the canonical closure contract requires proving the **median** residual <10%. The probe currently gates and records only `residual_max_pct`; this is a stricter bound when it passes, but the receipt does not explicitly publish the required median statistic. Please amend before merge so the receipt includes median residual and the gate tests that exact contract.

PR #1834, opened 2026-09-22
fix(ci): use server package for i1663 probe

gustavomhss:
Fixes the campaign package selector from the removed `corelink-container` package to the actual `corelink-server` package.\n\nRefs #1663.


gustavomhss:
Independent diff check: one line only, changing the suite selector to the actual `corelink-server` package. The prior hosted run 35686321546 failed because `corelink-container` was absent. Ready for required code owner approval.

gmhelmold:
Retriggering the repository-required DCO and rustfmt contexts for the unchanged bootstrap correction.

PR #1848, opened 2026-09-22
test(observability): bound read phase attribution

gustavomhss:
## Summary

- Add the read-path accounting invariant for the blocking CAS read and URL-map accounting window.
- Prove the two measured windows remain disjoint and their sum stays within request wall clock.
- This completes the focused B-122 regression coverage on top of the existing handle bridge and depth-controlled phase accounting.

## Linked work

- Work item / WI ID: B-122
- Spec ref: issue #1667
- Wave / sprint: observability
- Evidence: #1667 (implementation complete; production closure remains pending).

## Test plan

- [ ] Hosted focused campaign lane to be run after bootstrap.
- [x] Cold diff review: one test-only file, 8 added lines, no production or workflow changes.
- [x] `git diff --check origin/main...HEAD`

## Risk

- **Customer-visible change?** no — test assertion only.
- **Schema / API contract change?** no.
- **New crypto primitive / protocol?** no.
- **Worst-case if shipped tomorrow with one undiscovered bug?** The focused regression could be incomplete; production behavior is unchanged.

## Reviewer notes

The canonical `main` tree already contains both `spawn_blocking` ledger bridges, nested phase depth control, and worker timing consumers. This correction adds the missing read-side `Store + Accounting <= wall` assertion so overlap cannot regress silently.


gustavomhss:
Hosted campaign run [35688114931](https://github.com/HuGR-dev/corelink-server/actions/runs/35688114931) reached the exact adapter-cache lane but failed before the test because the workflow invokes the stale package ID `corelink-container`, which Cargo no longer exposes. This is workflow infrastructure, not a failure in the 8-line assertion; no production or workflow change is included in this PR.

gmhelmold:
Retriggering required hosted contexts before the authorized product merge.

PR #1763, opened 2026-09-22
feat(billing): add durable shadow aggregate state

gustavomhss:
## Context

WP #1630 adds the durable shadow runner aggregation state on top of the temporary public base `codex/i1639-public` (the #1639 stack). The base already owns the serialized D1 migration chain `0133`–`0137`; this stack assigns the new migration ordinal `0138`.

## What changed

- Add durable shadow aggregate state and cumulative shadow arithmetic.
- Preserve replay deduplication, partition invariance, and retroactive rate rejection.
- Add the bounded manual `i1630` GitHub Actions campaign suite.
- Add migration `0138_runner_aggregate_durable_state.sql` while retaining base migrations `0133`–`0137`.

## Criteria / DoD

- [x] Exclusive base: `codex/i1639-public`.
- [x] Migration chain `0133`–`0137` present from the base; #1630 migration is `0138`.
- [x] Manual `i1630` suite included in `.github/workflows/campaign-ci.yml`.
- [x] `git diff --check` clean.
- [x] Local CI/tests intentionally not run; this PR is prepared for GitHub Actions.
- [x] Refs #1630 (no closure).

## Risk / rollback

Shadow aggregation remains observational and bills nothing. The stack can be reverted as one PR if the hosted campaign exposes a defect.

## Review notes

Keep this PR stacked on `codex/i1639-public` until the #1639 base is integrated. Do not merge independently.

PR #1760, opened 2026-09-22
fix(billing): make Stripe entitlement authority durable

gustavomhss:
## Summary

- Reconcile runner entitlements from the current Stripe subscription authority.
- Wire the native Stripe authority through production container paths.
- Add a durable D1 CAS fence and assign migration ordinal `0137` after the base chain `0133`–`0136`.

## Linked work

- Work item: #1639
- Base stack: `codex/i1632-checkout-idempotency` (#1632)
- Refs: #1639

## Test plan

- [ ] Local CI/tests: intentionally not run; this PR is prepared for GitHub Actions.
- [ ] GitHub Actions: run on the public canonical repository after the stacked base is available.
- Mechanical checks completed: `git diff --check`; migration chain `0133`–`0137` present; working tree clean.

## Risk

- **Customer-visible change?** yes — billing entitlement reconciliation and production Stripe authority wiring.
- **Schema / API contract change?** yes — additive D1 migration `0137_runner_entitlement_reconcile_fence.sql`.
- **New crypto primitive / protocol?** no.
- **Worst-case if shipped tomorrow with one undiscovered bug?** A runner entitlement could be reconciled incorrectly; the durable authority key and CAS fence are intended to prevent stale Stripe state from regranting access.

## Reviewer notes

Please review the three WPs in order: current subscription reconciliation, native Stripe authority wiring, then the durable D1 CAS fence. The PR is intentionally based on `codex/i1632-checkout-idempotency` and should remain stacked until that base lands.

Issue #1630, opened 2026-09-17
[Audit][P2] Runner shadow charge reapplies the monthly allowance on every drain batch


gmhelmold:
## Evidence / scope
**Static producer/consumer proof plus an executed integer-arithmetic reference model.** The original Rust binary was not executed. This is a defect in the **shadow** calculation, not evidence that Stripe has undercharged a customer.

Baseline `cca798ff5bc2df660ecf2570ed243eb9775ff3d0`, inspected 2026-09-17.

## Source chain
- [Workflow drain and input construction](https://github.com/HuGR-Labs/corelink-server/blob/cca798ff5bc2df660ecf2570ed243eb9775ff3d0/.github/workflows/billing-aggregate-runner.yml#L220-L289) feeds only unaggregated rows (`runner_aggregated_at IS NULL`), bounded by `drain_limit`. The input carries prior hash-chain heads, but not prior monthly consumption.
- [Aggregator](https://github.com/HuGR-Labs/corelink-server/blob/cca798ff5bc2df660ecf2570ed243eb9775ff3d0/crates/corelink-runner-aggregate/src/lib.rs#L254-L398) constructs `tenant_totals` from only `input.staged`, then calls `overage_vcpu_seconds(*total,tier)` independently on that invocation's delta.
- [Money core](https://github.com/HuGR-Labs/corelink-server/blob/cca798ff5bc2df660ecf2570ed243eb9775ff3d0/crates/corelink-runner-overage/src/lib.rs#L123-L175) correctly computes `max(0,total-period_allowance)` for a **whole period**. The caller supplies a batch total instead.
- The workflow's later additive `runner_usage_counter` UPSERT preserves cumulative usage but does not feed that cumulative value back into the shadow computation that already ran.

Source blobs: aggregate `e2a0fb502a2f68129dc2e4c52cd6a9bfb2c9be62`; overage `277409f6f8df452e070e5633962c95bee1722e60`; workflow `112fdd2b3a975e63460ed670d028f30ad754dadb`.

## Counterexample
Keep one Starter tier unchanged for the entire month (allowance 100 vCPU-h). Drain 60 vCPU-h on day 1 and 60 vCPU-h on day 2.

```
Current two-run shadow: max(0,60-100) + max(0,60-100) = 0 h
Correct monthly excess: max(0,120-100) = 20 h
```

[...]

PR #1851, opened 2026-09-22
fix(db): serialize durable aggregate migration ordinal

gmhelmold:
## Summary

- Resolve the D1 migration ordinal collision by moving the #1630 durable aggregate migration from `0138` to `0139`.
- Keep `0134_runner_aggregate_durable_state.sql` as the authoritative migration fixture path.
- Preserve the serialized `0133`–`0137` queue and update the billing workflow reference to describe `0139` as the compatibility no-op.

## Stack

- Base: #1760 (`codex/i1639-public`)
- Dependent: #1763 (`codex/i1630-public`)
- This correction is based on the current #1763 head.

## Linked work

Refs #1845
Refs #1630

## Validation

- `git diff --check` passed.
- Confirmed the branch is based on the current `codex/i1630-public` head.
- Local tests/builds intentionally not run; use the required GitHub-hosted migration validation after the stacked base is available.

## Risk

Migration file rename and reference correction only. No destructive SQL or existing merged migration rewrite.


gmhelmold:
Hosted validation evidence:

- Run: https://github.com/HuGR-dev/corelink-server/actions/runs/35688979693
- Head: `e71135905cefa12b9e0a40811e0b1a91e58fc6ae`
- Workflow: `campaign-ci`, suite `i1630`
- Mandatory pytest bootstrap completed successfully (`pytest==8.3.5`).
- The suite failed on inherited stack state: `cargo fmt --all --check` reported existing diffs under `crates/corelink-container`, and the aggregate crate test reported an existing assertion failure in `crates/corelink-runner-aggregate/src/lib.rs:1154`.

No retry was dispatched because the failure requires a separate dependent-stack correction. The migration ordinal change and its workflow/fixture references were not the failing commands.

PR #1858, opened 2026-09-22
fix(ci): install protoc for i1663 campaign lane

gustavomhss:
The corrected i1663 selector now reaches Cargo, but the hosted lane fails before the focused test because setup-protoc excludes suite i1663.

This one-line condition addition makes the existing bounded lane executable. Refs #1663, #1667.


gustavomhss:
Retriggering required hosted contexts for this one-line campaign lane correction.

Issue #1669, opened 2026-09-19
[B-127] 3 670 audit rows have no tenant, so the residency predicate is unevaluable — and a JOIN check reports clean


gmhelmold:
## Success criteria

- #1669 / B-127 remains a P1 security go-live blocker for `corelink-server`; it is not cosmetic. A residency result is valid only when it classifies the full `audit_outbox` population with a `LEFT JOIN`, including a failing `unevaluable` state for rows that cannot resolve a tenant.
- The closure gate is a fresh, reproducible, redacted production receipt whose aggregate population reconciles and is `COMPLIANT`, plus an authority-backed disposition for every historical unexplained orphan. Repository prose, a source-only test, an inner-join result, or unrelated green CI does not satisfy that gate.
- Current repository state was audited on `main` at `a706ae1c9ff4c4ff2f1bc2569081b1c95fc0eb96`: the preventive writer and tenant-delete guards, full-population verifier, bounded read-only probe, and offline aggregate classifier are present. The pertinent B-127 assets are unchanged since merged PR #2117.
- The latest linked production receipt is hosted run [35697251287](https://github.com/HuGR-dev/corelink-server/actions/runs/35697251287), captured from `main` at `58228ac21f3b92a49e091a791bf41bb2c77f4c87`. It is fail-closed: 84,985 total rows, 81,315 satisfied, 0 violated, 3,539 unevaluable/orphan, 3,526 erased-orphan retained rows, 13 unexplained rows across 4 tenants, 131 valid `_public` rows, and 0 invalid `_public` rows. Its receipt status is `FAILED`.

## Definition of done

- The restricted data owner reconciles the four tenant identities and records an evidence-backed disposition for all 13 unexplained retained rows. The 3,526 confirmed Art. 5(2) retained audit rows remain preserved.
- If a historical repair is required, it is handled by a separately approved, bounded, auditable workflow with the required data-owner/legal approval; it must not be inferred from this issue or performed by the read-only lane.
- A fresh production read through the approved restricted evidence path produces a redacted aggregate receipt that passes the full-population, lineage, and checksum checks. Any repository implementation is merged and verified on the exact resulting `main` tree.
- The issue links the fresh evidence with secrets and personal data redacted. `BACKLOG.md` is reconciled only in the actual closing change, and any package-created PR, branch, worktree, or follow-up is owned or removed.

## Completeness criteria

[...]

PR #1878, opened 2026-09-22
fix(audit): reconcile public namespace in 1669 probe

gustavomhss:
## Summary

- Aligns the issue #1669 backfill aggregate with residency's explicit `_public` system namespace partitions.
- Keeps valid and invalid `_public` rows distinct from customer orphan rows.
- Preserves aggregate-only counts in receipts when reconciliation fails, so a mismatch remains classifiable without row or tenant data.

## Linked work

- Work item: #1669 / B-127
- Spec ref: `reports/go-live/D-2-residency-third-state.md`
- Closes: none; fresh hosted evidence is still required after merge.

## Test plan

- `python3 -m pytest -q tests/test_probe_i1669_readonly.py` — 3 passed.
- Fixtures cover satisfied, cross-region, erased orphan, unexplained orphan, valid `_public`, and invalid `_public` rows.
- No live D1 query or mutation is included.

## Risk

- Customer-visible change? no.
- Schema / API contract change? no; this changes only the read-only evidence query and its redacted aggregate receipt.
- Worst-case if shipped tomorrow with one undiscovered bug? The bounded probe remains fail-closed and reports `INDETERMINATE` rather than declaring residency compliant.

Refs #1669


gustavomhss:
Hosted credentialless contract evidence:

- Run [35693947220](https://github.com/HuGR-dev/corelink-server/actions/runs/35693947220) passed on `ubuntu-24.04` against head `b206824adbe3e2255fccc2aff121d1c6736c9884`.
- The i1669 campaign step ran the existing writer checks plus `tests/test_probe_i1669_readonly.py`; all three adversarial fixtures passed.
- No live D1 query, credential, backfill, or mutation was used.
- The earlier run [35693678278](https://github.com/HuGR-dev/corelink-server/actions/runs/35693678278) failed because the fixture schema predated current `event_type` and reserved-public residency state; that was corrected in `b206824ad` and the follow-up hosted run is green.

Both temporarily enabled workflows have been disabled again. The live issue-1669 read-only workflow was not dispatched.

PR #1853, opened 2026-09-22
fix(billing): fence replaced subscription identities

gmhelmold:
## Summary

- Fence Runner entitlement mutations by both authority key and subscription identity.
- Permit a legitimate successor grant to replace a predecessor revoke while rejecting predecessor cancellation after successor grant.
- Preserve migration 0137 and add migration 0138 for the new `is_granting` column so already-applied 0137 schemas remain compatible.

## Linked work

- Closes #1844
- Refs #1639
- Follow-up to #1760

## Test plan

- [ ] Local tests/builds: intentionally not run per task constraint.
- [ ] Hosted campaign: dispatch the canonical `billing campaign CI` `i1639` suite at this head.
- [x] Read-only checks: `git diff --check`; DCO trailers preserved on every commit.
- [x] Migration review: 0137 remains unchanged; 0138 is additive with a safe default for historical fence rows.

## Risk

- **Customer-visible change?** yes — stale Stripe subscription snapshots can no longer revoke a successor Runner entitlement.
- **Schema / API contract change?** yes — additive migration 0138 extends the existing durable fence.
- **New crypto primitive / protocol?** no.
- **Worst-case if shipped with an undiscovered bug?** A stale Runner webhook is rejected for retry, leaving the durable fence and entitlement unchanged.

## Reviewer notes

The key acceptance cases are successor grant after predecessor revoke, predecessor revoke after successor grant, equal-period identity retries, and tenant-scoped SQL CAS predicates. No Stripe writes or live activation are involved.

PR #1896, opened 2026-09-22
fix(billing): make Stripe entitlement authority durable

### LINKED ISSUE #1639: [Audit-R4][P2] Native runner entitlement follows webhook arrival order rather than current subscription state
## Evidence and scope
**Reproduced with the original compiled Rust materializer**, on Intel macOS/Rust 1.91.1, using its real `StateMaterializer::on_subscription_updated` implementation and the existing in-memory D1/audit seams. Baseline `cca798ff5bc2df660ecf2570ed243eb9775ff3d0`, inspected and tested 2026-09-17. Production source was unchanged. No live Stripe event, account, subscription, HTTP ingress or D1 service was used.

Proposed P2: a path-specific convergence defect. This is not a statement that a production customer has lost access or that every webhook authority behaves identically.

## Source and external contract
- [`do_subscription_upsert` and `reconcile_runners`](https://github.com/HuGR-Labs/corelink-server/blob/cca798ff5bc2df660ecf2570ed243eb9775ff3d0/crates/corelink-billing-stripe-materializer/src/handler.rs#L324-L539) act on the delivered subscription snapshot. An `active`/`trialing` runner price seeds entitlement; a non-granting status deletes it. This path does not establish that the snapshot is still the current state of the purchased subscription.
- [Production writer SQL](https://github.com/HuGR-Labs/corelink-server/blob/cca798ff5bc2df660ecf2570ed243eb9775ff3d0/crates/corelink-billing-stripe-materializer/src/d1.rs#L263-L282): the runner UPSERT and DELETE are keyed by tenant and carry no subscription-state/revision precondition. A cache-axis cancel guard elsewhere is not a runner ordering guard.
- [Native composition](https://github.com/HuGR-Labs/corelink-server/blob/cca798ff5bc2df660ecf2570ed243eb9775ff3d0/crates/corelink-container/src/main.rs#L778-L876) wires this handler as a second entitlement writer when its configuration is present. Calling it defense-in-depth does not prevent it from overwriting a newer projection made by another writer.
- Stripe's [webhook delivery documentation](https://docs.stripe.com/webhooks#event-ordering) does not guarantee generation order. Its current guidance also warns against treating `created` as a total ordering or deduplication key; resource reconciliation and event identity are separate concerns.

## Executed counterexample and control
One synthetic tenant has a recognized runner price and one subscription, with a deliberately known business sequence: `past_due` followed by `active`. Two distinct event IDs represent those snapshots. Their synthetic `created` values establish the test scenario, not a proposed ordering algorithm.

Deliver the newer active snapshot first, then the older past-due snapshot:
```
new active event -> runners entitlement Some((40, 240))
old past_due event -> runners entitlement None
R4_EVENT_ORDER final_entitlement=None
```
`audit_r4_old_runner_status_must_not_override_newer_active_event` fails its desired final-state assertion. Deliver exactly those states in chronological order as a control: entitlement remains `Some((40, 240))`, and the independent Cache downgrade recorder is empty. That control passes.

Both deliveries are distinct events. Fixing #1629's duplicate/crash ownership problem alone cannot make the stale distinct event disappear. The test does not depend on a duplicate event ID, storage failure, changed tier, invalid signature, or a price calculation.

## Impact and boundary
The final runner entitlement depends on network delivery order rather than the latest business state in this materializer. An old non-granting event can revoke recovered service. Conversely, other reordered granting events require review for unintended re-grants; that opposite direction is not claimed as an executed case here.

The in-memory writer was exercised; the matching unconditional production SQL was inspected, not executed against Cloudflare. Any independent repair sweep might eventually correct the projection, but none is consulted by this handler before its mutation, and it would not make this write correct at that time.

## Focused correction
Treat a delivered event as an idempotent reconciliation trigger for the identified subscription/purchase, then obtain or prove the authoritative current entitlement under a defined concurrency rule. Reuse existing reconciliation and purchase mappings where available. Ensure the primary and secondary writers cannot overwrite one another with stale snapshots.

Do not simply sort or gate solely by `event.created`: equal-second events and lack of a provider total-order guarantee remain. Do not rely on event-ID lexicographic order. Fetching current state also needs protection against an older reconciliation completing after a newer one. A general event-sourcing rewrite is not required; define the smallest shared authority that makes the projection converge.

## Success criteria
For the same actual subscription lifecycle, supported delivery permutations and retries converge to the same current runner entitlement. A stale distinct event cannot undo the recovered active state.

## Completeness criteria
Cover granting/non-granting updates, cancellation, changed/replaced subscription identities, retries, primary/secondary writer interaction, reconciliation failures and restart. Keep Cache and Runner axes separate. Coordinate with #1629 without conflating event ownership with state freshness.

## Quality standards
Keep the original-code red test and chronological positive control. Add an explicit current-subscription oracle to the repaired integration, test both delivery orders and equal-created timestamps, and synchronize two reconciliation attempts to finish in reverse order. Assert persisted entitlement and purchase identity, not only returned success. Exercise the actual writer composition before claiming production-equivalent acceptance.

## Definition of done
The red original-code regression becomes green under the chosen current-state authority; the chronological/duplicate/failure controls remain correct; any existing stale projections have an explicit reconciliation procedure. Record source identities, commands and observed final state. No Stripe writes, price changes or live activation are authorized by this issue.

## Invariants
Arrival order is not business order. An event ID is not a subscription revision. A stale second writer cannot override authoritative current entitlement. Temporary inability to reconcile remains owned and observable. Cache entitlement and signature checks are unchanged.

## Duplicate review
Repository-scoped `out-of-order` search found #1629 mentioning ordering only in its test matrix. Its demonstrated root cause is dedup-before-effect, whereas this issue uses two different events and no crash. No separate matching state-freshness issue was found. Do not interpret this targeted review as a proof that every historical note lacks a related observation.

### LINKED ISSUE #1844: [Follow-up] Runner entitlement fence must reject replaced subscription identities
## Context

Cold review of PR #1760 (issue #1639) found that the durable Runner entitlement fence does not handle replaced subscription identities.

## Finding

The fence row stores `stripe_subscription_id`, but `SQL_ADVANCE_RUNNER_ENTITLEMENT_FENCE` compares only `authority_key` and updates the subscription id only when the key is lexicographically newer. The native key is built from `current_period_end`, status rank, and `price_id`.

For a tenant whose subscription is replaced within the same billing period, an old subscription's `canceled` snapshot (rank 80) can compare newer than the replacement subscription's `active` snapshot (rank 10). The old subscription can therefore revoke the replacement entitlement. The stored subscription id is not used as a lifecycle guard.

PR #1760 tests cover one subscription id and do not cover replacement or changed subscription identity. Issue #1639's completeness criteria require this case.

## Required correction

Define and test a tenant-level authority that rejects stale snapshots from a replaced subscription, including:

- replacement subscription active followed by old subscription canceled;
- reverse delivery and equal-period cases;
- persisted entitlement and purchase identity assertions;
- production D1 CAS behavior, not only the in-memory mirror.

No Stripe writes or live activation are authorized by this follow-up.

Refs #1639
Related PR #1760

gustavomhss:
## Summary

- Rebase the #1639 entitlement convergence stack onto current `main` and include the #1844 replacement-identity correction.
- Fetch the current Stripe subscription state before Runner mutation and fence the durable D1 entitlement projection by tenant, subscription identity, authority key, and grant state.
- Keep migration 0137 immutable and extend it through additive migration 0138.

## Correctness coverage

- Out-of-order granting/non-granting deliveries reconcile from current state.
- Replays converge without changing the durable entitlement.
- A successor grant replaces a predecessor revoke; a predecessor cancellation cannot revoke the successor identity.
- CAS writes fence and entitlement mutation in one D1 batch, and the retained fence survives revocation/restart.
- Migration 0138 adds `is_granting` with a safe default for already-applied 0137 rows.

## Validation

- [x] `git diff --check`
- [x] SSH-signed commits with DCO trailers
- [ ] One hosted `billing campaign CI` run with suite `i1639` at this exact head
- [ ] Hosted DCO and rustfmt checks

## Risk

A stale Runner snapshot is rejected for retry without changing the durable fence or entitlement. This change makes no Stripe writes and does not change the Cache entitlement axis.

Closes #1639
Closes #1844
Supersedes #1760 and #1853


gustavomhss:
## Merge blocker: cold review findings

This PR is draft and must not merge. The combined #1639 / #1844 stack still has three correctness gaps:

1. **P0 — synthetic authority key is not a provider revision.** `current_period_end + status rank + lexical price id` is used as a durable ordering key. In one period, a recovered `past_due` subscription can become `active` with a lower status rank; the active current snapshot is then classified stale and the entitlement remains revoked. A price replacement whose id sorts lower has the same failure mode.

2. **P0 — a successor cancellation is blocked by the replacement fence.** With `sub_A` currently granting, a known replacement `sub_B` that is already canceled cannot take ownership because cross-identity CAS only permits an incoming grant after a stored revoke. The handler retries the `sub_B` cancel while retaining `sub_A` entitlement. The existing identity test only proves predecessor cancel after successor grant, not this direction.

3. **P1 — the fetched current snapshot is not written to `stripe_subscriptions`.** The entitlement path reads current Stripe state, but the durable subscription projection still stores delivered webhook status. A stale active replay can therefore overwrite a fetched current canceled status in that projection.

The first focused run passed materializer and authority tests but failed server Clippy on these new test fixture unwraps plus inherited dead-code findings: https://github.com/HuGR-dev/corelink-server/actions/runs/35695335354

The bounded correction removed the test-only Clippy errors, rebased onto current main, and dispatched the final permitted focused run at `c7c468657bdfe534f8251187ae548cb84fa33b5a`: https://github.com/HuGR-dev/corelink-server/actions/runs/35695971230

DCO, rustfmt, and Docs Reality Gate are green on the final head. No further correction is authorized in this review cycle. #1639 and #1844 remain open.

gustavomhss:
Closing this draft as superseded by canonical draft PR #1906.

Evidence:
- PR #1906 explicitly supersedes the draft authority-key design in #1896.
- #1906 replaces the synthetic period/status/price ordering key with provider subscription creation time plus signed event creation revision.
- #1906 includes the replacement-identity and successor-cancellation path from #1844, current-subscription persistence, and the provider-backed CAS fence.

#1906 remains FIX-FIRST and is not being merged: the independent signup-worker Stripe path still writes `runners_entitlement` without the shared durable fence (P0 stale-second-writer invariant), and equal provider creation timestamps need an explicit deterministic policy and acceptance coverage. Issue #1639 and #1844 remain open.

PR #1906, opened 2026-09-22
fix(billing): use Stripe revisions for runner authority

### LINKED ISSUE #1639: [Audit-R4][P2] Native runner entitlement follows webhook arrival order rather than current subscription state

gustavomhss:
## Summary

- Reconcile Runners entitlement from a fetched current Stripe subscription, then persist that fetched projection in `stripe_subscriptions`.
- Replace the rejected synthetic key with numeric provider coordinates: `subscription.created` selects replacement identity and signed `event.created` orders revisions of that identity.
- Fence the durable D1 entitlement mutation with the projection identity, including successor cancellation and stale predecessor safety.

## Linked work

- Work item / WI ID: #1639
- Closes: #1639
- Supersedes the draft #1896 authority-key design.

## Test plan

- [ ] Hosted `billing campaign CI` suite `i1639` at this head.
- [ ] Hosted DCO and rustfmt PR checks.
- [ ] Independent cold review of same-period recovery, price replacement, successor cancellation, predecessor replay, out-of-order delivery, and fetched-current `stripe_subscriptions` persistence.

## Risk

- **Schema / API contract change?** yes — additive migration `0137` creates the durable fence and records provider revision metadata.
- **Worst-case if shipped tomorrow with one undiscovered bug?** a webhook remains retryable rather than applying an entitlement from non-provider ordering data.

## Reviewer notes

- Price IDs choose an entitlement ladder only. They never establish authority.
- A replacement’s non-granting current state may own the fence and revoke access; an older identity cannot subsequently mutate it.


gustavomhss:
## FIX-FIRST cold audit (canonical path)

PR #1906 is the selected canonical draft after #1896 was closed as superseded. It addresses the synthetic authority-key defects by using provider subscription creation time and event revision, persists the fetched current projection, and covers predecessor/successor delivery order.

P0 blockers before merge:

1. The signup-worker remains an independent Stripe writer of `runners_entitlement` (`apps/signup-worker/src/webhooks/stripe_persistence_runner.ts`, direct INSERT/DELETE). It does not participate in `runner_entitlement_reconcile_fence` or the provider-current authority. A stale signup-worker delivery can still overwrite or delete a projection after the native container CAS, violating the invariant that a stale second writer cannot override authoritative current entitlement and the required primary/secondary writer composition coverage.
2. Provider subscription creation timestamps are stored as milliseconds derived from Stripe seconds. Distinct replacement subscriptions created in the same provider second compare equal; the cross-identity CAS then has no deterministic ownership rule. The acceptance contract needs an explicit equal-coordinate policy and tests.

Hosted evidence:
- The prior focused `i1639` run for #1906 failed before acceptance due unrelated repository-wide Clippy/dead-code findings in `corelink-server` (`tier_select/part-00-01.rs`, `tier_select_store.rs`, and inherited byte-accounting helpers): https://github.com/HuGR-dev/corelink-server/actions/runs/35697544040
- DCO, rustfmt, and Docs Reality were green on the prior head, but they do not clear the P0 blockers.

No merge requested; issues #1639 and #1844 remain open.

Issue #1659, opened 2026-09-19
[B-103] o caminho de escrita do `/cargo` falha em 87% sob paralelismo e satura em ~2 req/s


gmhelmold:
## Success criteria

- Repo code removes avoidable false serialization between independent `/cargo` CAS writes at the historical 220-write population by enlarging only the fixed CAS accounting lock table.
- Same `(tenant, content_hash)` write/delete accounting remains serialized; atomic quota checks and reserve-before-store behavior remain intact.
- The current repo correction does not claim the historic production rate, 429 attribution, or runtime acceptance. Those require the protected live receipt below.

## Definition of done

- The exact-head GitHub-hosted repository checks, DCO, and rustfmt pass; a single consolidated Luna cold review approves; root merges the canonical PR; main is checked at the merged SHA.
- The external runtime receipt is linked and `BACKLOG.md` reconciled before this issue closes.

### Ownership ledger

- Owner: `/root/cargo_write_1659` (central-orchestrator delegate)
- Branch: `fix/i1659-cas-write-shard-contention-20260924`
- Head: `e6ceaf8254272c443ec54c40ec62c0ea63be5730` (signed+DCO). PR base: `main@24e8d41c70e581b0759f3a7347ea417fd4bc7185`; branch merge-base: `2af5c9b64cccba9cf618145224186ec2717cca50` (main is one commit ahead of the branch at this audit).
- PR: #2407 (draft, canonical); comparison to current `main` is diverged by one commit on each side, with merge-base `2af5c9b64cccba9cf618145224186ec2717cca50`; #2147 remains closed and historical, superseded by this source-only correction; #2147 remains closed and historical, superseded by this source-only correction
- Hosted runs: `i1659-dry` run [35955979325](https://github.com/HuGR-dev/corelink-server/actions/runs/35955979325), success at exact head `e6ceaf8254272c443ec54c40ec62c0ea63be5730`; both the B-103 dry reproducer contract and CAS write-lock regression jobs succeeded. DCO, rustfmt, B-102 cap propagation, and CodeQL also completed successfully at this head. Campaign workflow enabled, manual-only.
- Cold review: Luna APPROVE at exact head `e6ceaf8254272c443ec54c40ec62c0ea63be5730`, merge-base `2af5c9b64cccba9cf618145224186ec2717cca50`; final hosted proof passed in run `35955979325`; CodeQL completed successfully.

## Completeness criteria

[...]

PR #1992, opened 2026-09-22
test(ci): bind B-103 dry gate to warm writes

gustavomhss:
## Summary

Bind the credentialless `i1659-dry` hosted contract to the current B-103 reproducer semantics.

The merged reproducer now requires:

- three sequential warm PUTs before the concurrency arms;
- unique per-key payloads so the matrix cannot synthesize same-blob CAS serialization;
- redacted request body digests;
- warm failures included in the fail-closed verdict.

The dry gate previously checked only the older concurrency and method markers. This change makes removal or regression of the warm phase and unique payload contract fail the focused hosted gate.

## Scope and safety

- One workflow contract file; no production code, credentials, staging requests, or production mutation.
- The separate owner-dispatched staging workflow remains the only runtime load path and remains unreachable from this dry gate.
- No throughput or live behavior claim is made.

## Verification

- Commit is signed and includes the required DCO trailer.
- Local tests/build/lint were intentionally not run; the focused `i1659-dry` gate must run on GitHub Actions.
- Runtime closure still requires the authorized staging receipt, three hot sequential PUTs, bounded parallel population, unfiltered tail evidence, and redacted wire evidence described in #1659.

Refs #1659


gustavomhss:
## Cold review verdict: FIX-FIRST

The hosted focus, scope, nonclaims, contamination boundary, and DCO are sound:

- the job is selected only by the closed `i1659-dry` input and checks out the dispatched revision;
- the dry lane is credentialless/static and explicitly does not issue staging requests;
- the PR changes only `.github/workflows/campaign-ci.yml`;
- the exact head is signed and the DCO check is green.

One material completeness gap remains in the mutation sensitivity of the B-103 contract. The new assertions are source substring markers plus top-level function-name checks; they do not bind the warm implementation to the declared three PUTs or bind the digest marker to the PUT response path. For example, these mutants would retain every current marker and pass the dry gate:

1. change `for _ in range(3)` in `run_warm_sequence` to `range(2)` while leaving the artifact's `"warm_sequence": ["PUT", "PUT", "PUT"]` declaration unchanged;
2. change the warm `request(..., "PUT", key, body)` call to another method or omit the body while retaining the existing `body` and `request_body_sha256` strings elsewhere in the file.

Because #1659 explicitly requires the warm sequence, unique payload, and request digest contract to be mutation-sensitive, the dry gate needs an AST/semantic assertion or equivalent executable mutation proof that these wiring changes fail. Until then the gate can report PASS while the runtime reproducer no longer executes the required warm PUT matrix.

PR #2001, opened 2026-09-22
fix(B-154): restore fail-closed claim guard

gustavomhss:
## Context

B-154 / issue #1676 remains owner-blocked. The executed DPA Object Lock claim and Enterprise SLA BYOK p99 claim lack current operational proof; the case study remains a restricted draft.

The immutable v0003 ledger receipt correctly records its own historical source, but later `main` backlog drift was never recorded. That made a normal v0004 successor reject its true BASE preimage. The repository census now also truthfully reports 486 full-schema specifications plus 11 YAML-only specifications, 497 total.

## What

- Add one byte-pinned v0004 reconciliation bridge from `main@0d1e85792bbe1b495bc8273ee63011417510140f`.
- Preserve v0003 and every catalog byte; accept only the exact B-098 census correction and B-154 owner-packet-and-claim-verifier conjunction.
- Restore the B-154 fail-closed verifier and update it for the current schema-v2 blocked B-083 receipt.
- Update the B-098 prose, `CLAUDE.md`, and redacted blocker receipt together.
- Keep B-154 open and preserve unresolved DPA, SLA, and case-study decisions.

## How tested

Hosted GitHub Actions only. The focused workflow runs the canonical B-154 self-test and byte-pinned ledger mutation suite; the existing B-098 credentialless workflow validates the census and blocker receipt.

## Risk / rollback

The bridge is bound to one sequence, predecessor, base commit, complete backlog hashes, ledger hashes, catalog hashes, changed IDs, and exact B-154 verifier command. Revert the signed PR commits to remove the proposed successor; no legal instrument, owner receipt, or external decision is altered.

Refs #1676 #1657


gustavomhss:
Closed because its required BASE-owned policy predecessor PR #2023 was rejected after the final bounded correction. Merging this data successor without a trusted policy in main would preserve the self-authorization flaw identified by cold review. No further correction loop will be attempted in this PR.

Issue #1668, opened 2026-09-19
[B-125] the audit chain seals 200 rows/hour, so evidence lags the event by hours


gmhelmold:
## Classification

- Priority: **P1**
- Type: **correctness**
- Delivery impact: **blocks-go-live**
- Readiness: **needs-evidence**
- Repository: **corelink-server**
- Cosmetic: **no**

## Dependency graph

- Blocked by: None.
- Unlocks: None.
- Related, not blocking: None.

**Unblock rule:** Unblocks/finishes on the fresh live or authority evidence named in the acceptance gate; repository prose alone is not proof.

## Scope

Measured against production D1 (`d64742ea`) on 2026-08-30, with a control query
beside every count so a zero means absence and not a broken instrument. Full
method in `reports/go-live/D-1-audit-trail.md`.

| measure | value |
|---|---|
| rows sealed per hour, six consecutive hours | **200, 200, 200, 200, 200, 200** |
| seal latency, last 24 h (n=2692) | min **12 s**, mean **1 h 28 min**, max **5 h 13 min** |
| unsealed backlog | **4 243** of 77 935 |
| oldest unsealed row | **6.8 h** |
| arrivals over the same window | 86, 300, 2, 30, 88, 92 (mean ≈ 100/h) |

Six identical integers is a hard cap, not a load curve. The drain currently
outpaces the mean arrival rate, so the backlog shrinks — but the entire margin is
a factor of two against a bursty process whose **observed peak of 300/h already
exceeds the ceiling**. A sustained burst accumulates and then drains only at
200/h.

**What the customer gets:** an event is not tamper-evident when it happens. It
becomes so ~1.5 h later on average, and in the observed tail **over 5 h** later.
Anything read from `audit_outbox` inside that window is present but outside the
chain.

**Not decided here, deliberately:** where the 200 comes from. It is not a literal
in `routes/audit_drain.rs` — grep returned 0 there against a control of 5 hits
for another string in the same file, so the absence is real; the value is passed
by the caller and the caller was not identified. Guessing it would be the kind of
claim this campaign keeps retracting.

**What IS settled and should stop being re-litigated:** the chain head signing is
correct and live — 365 of 365 heads carry an Ed25519 signature, and the seed is a
**write-only Cloudflare secret** (`ERASURE_ATTESTATION_SEED_HEX`, reused by
design), with **zero** occurrences in `wrangler.toml` against a control of 5. An
[...]

PR #2014, opened 2026-09-22
fix(b125): retain redacted provider diagnostics

gustavomhss:
## Summary

- retain bounded, credential-redacted Wrangler/D1 stderr diagnostics in the B-125 read-only receipt
- preserve fail-closed query semantics and remote SELECT-only guards
- add static mutation checks and a credentialless focused contract workflow

Refs #1668

## Verification

- `git diff --check` passed
- signed commit: `2b6725beb2c9c2bd1ad2d2ca96aa6eed2c60aa8c`
- local tests/build/lint not run per task constraint
- focused hosted contract workflow runs the static contract checks


gustavomhss:
Rejected after the bounded cold re-review: the scrubber still leaks a standalone `Bearer <secret>` form. The branch is not merged; no diagnostic output change reached `main`. A future replacement must begin from a clean threat model and cover standalone bearer plus prefixed provider-token forms in executable tests.

PR #2023, opened 2026-09-22
fix(ledger): install trusted v0004 reconciliation policy

gustavomhss:
## Summary

Installs the trusted, exact byte-pinned v0004 reconciliation policy before any
successor data is proposed for #1676.

This PR changes no `BACKLOG.md`, ledger receipt, snapshot, catalog, CLAUDE
census, or owner-action evidence. It installs only the policy controls:

- exact v0004 prestate, target state, catalog, receipt-field, and B-154
  command/means pins;
- the fail-closed B-154 canonical verifier and focused tests;
- a hosted end-to-end proof that simulates policy already in BASE and accepts
  only the exact v0004 candidate bound to that immutable BASE event SHA.

`base_commit` deliberately uses the generic `receipt.base_commit == trusted
BASE HEAD` rule. The pinned prestate and unchanged catalog bytes prevent this
from authorizing another BASE state. The policy authorizes precisely sequence 4
with B-098 and B-154 together; it adds no generic exception.

After this merges, the data-only successor can be checked by the existing
BASE-owned `backlog-verify` workflow.

Refs #1676 #1657


gustavomhss:
Rejected after the final bounded correction. Exact head 0b7898b9eef2ee254beaadc100d251875a7ef152 fails current hosted policy run 35766674795 deterministically: the end-to-end base-policy mutation case expected the trusted-BASE rejection but hit delivered-state drift instead. This is neither stale nor infrastructure. Per the campaign no-loop rule, no further correction will be attempted in this PR.

PR #2026, opened 2026-09-22
fix(gc): account purge by fenced gc region

gustavomhss:
## Context

Closes the repo-owned B-071 accounting seam. `blob_meta.region` is a macro-residency value (`wnam`, `enam`, etc.), while `tenant_storage_state.region` is the five-region GC partition (`iad`, `lhr`, `nrt`, `sam`, `syd`). The historic deployed 0118 trigger bodies compared those different domains via `OLD.region`, so non-`sam` purges could fail the accounting-row guard and could not reclaim usage.

## What

- Correct 0142's accounting guards to select the fenced purge intent's `gc_region`.
- Add migration 0143, which replaces the two accounting trigger bodies already installed by historic 0118 deployments; legal-hold and atomic finalization guards remain intact.
- Exercise historic 0118 -> 0143 replay twice, plus the 0142 -> 0143 fresh-install path, in the credentialless hosted verifier.

## How tested

- No local test/build/lint was run.
- Hosted B-071, D1 validation, DCO, and rustfmt were triggered for restacked head `8f796a2e5f22ac703c3f277ad008c1112cc339b2`, based exactly on `589deab43b85c13c2896a88d5b55f545d640b65b`.

## Invariants

- The purge intent's `gc_region` is the sole accounting partition selector.
- Legal hold, tenant/digest fencing, atomic audit/candidate/intent updates, saturated byte accounting, and idempotent replay remain fail-closed.

## Risk / rollback

0143 only replaces the two obsolete accounting trigger definitions. Revert before migration deployment if rollback is required.

Refs #1651


gustavomhss:
Rejected after the bounded integration correction. Exact head 8f796a2e5f22ac703c3f277ad008c1112cc339b2 passes B-071 and unique-prefix checks, but hosted D1 run 35768965729 correctly rejects the two DROP TRIGGER statements under INV-AUTH-MIGRATION-ADDITIVE. Further work is split into #2033 with an explicit name-scoped ADR/policy contract; no second correction will be attempted in this PR.

Issue #2033, opened 2026-09-22
[B-071/D1] Authorize a name-scoped forward trigger replacement migration


gustavomhss:
## Problem

PR #2026 proved the B-071 accounting fix but its forward migration must replace two already-deployed SQLite triggers. The repository-wide additive migration gate rejects `DROP TRIGGER IF EXISTS`, so the safe upgrade cannot merge without an explicit, bounded policy for trigger replacement.

## Success criteria

- Define a fail-closed, ADR-approved exception limited to replacing the two named B-071 accounting triggers.
- Preserve atomic accounting, legal hold, tenant/digest fencing, concurrency, idempotence, and saturated byte arithmetic.
- Prove both fresh install and historical `0118` -> ordered migration upgrade paths.
- No generic destructive-DDL bypass.

## DoD / completeness

- One signed+DCO PR based on current main after #2029.
- Migration order remains `0142` base, `0143` upgrade.
- Additive verifier accepts only the exact named trigger replacement and rejects table/index/column drops or any other trigger.
- Hosted D1, B-071, mutation, rustfmt, and DCO checks green on exact head.

## Invariants / quality

- No production apply from the PR.
- No weakening of global additive migration policy.
- Exception is byte/name scoped, documented in ADR, and mutation tested.
- One bounded correction cycle.

Refs #1651 and supersedes rejected PR #2026.

PR #2036, opened 2026-09-22
fix(d1): scope B-071 GC trigger replacement

gustavomhss:
Refs #2033 and #1651. Success: ADR-0143 constrains 0143 to two named accounting-trigger replacements; 0142 uses fenced gc_region; fresh and historical replay plus destructive-DDL mutations are covered. DoD: signed DCO commits, no production apply, hosted D1/B-071/fmt/DCO checks requested. Invariants: legal hold, tenant/digest fencing, atomic accounting, idempotence, and saturated bytes remain fail-closed.


gustavomhss:
Rejected after the final bounded correction. Exact head 080de720 passes hosted gates, but cold review found the Rust waiver searches the whole migration for an approved DROP and therefore also permits a later unpaired duplicate CREATE of an allowed trigger name. This violates #2033 exact-pair authorization. No further correction will be attempted in this PR; #2033 remains open.

Issue #1720, opened 2026-09-20
[P1] Corrigir contrato falso de OIDC no terraform-drift


gmhelmold:
## Problema

`.github/workflows/terraform-drift.yml` declarava Cloudflare via GitHub OIDC e concedia `id-token: write`, mas não solicitava nem trocava um ID token. Também passava `CF_CLIENT_ID`/`CF_CLIENT_SECRET` diretamente ao provider pinado, que autentica com credenciais Cloudflare próprias. A claim de OIDC não correspondia ao comportamento.

## Escopo

- usar um segredo dedicado `CF_TERRAFORM_DRIFT_API_TOKEN`, sem reutilizar o token amplo de deploy;
- mapear as variáveis suportadas pelo provider e a variável não secreta `CF_ZONE_ID`;
- remover `id-token: write` e as claims OIDC falsas;
- exigir exatamente seis permissões read-only: account `Workers R2 Storage Read`, `D1 Read`, `Workers KV Storage Read`, `Workers Scripts Read`; zone `DNS Read`, `Workers Routes Read` limitada a `humangr.com`;
- proibir Edit/Write, API-token management, Access, Pages e user scopes;
- preservar `terraform plan` sem `apply`, propagação do exit code e ausência de segredo em logs.

## Aceite

- workflow, provider e inventário usam nomes coerentes;
- token compartilhado não é consumido pelo drift;
- `CF_ZONE_ID` é injetado por repository variable e inputs ausentes falham loudly;
- testes mutation-sensitive rejeitam remoção do exit final e qualquer sétima permissão;
- actionlint, testes focais e diff check locais passam;
- cold review aprova;
- antes do merge, o token dedicado é provisionado como repository secret, validado por leituras dos seis recursos e nunca exibido.

## Dependências e follow-ups

- Encontrado durante #1702, mas não bloqueia a transferência: o deploy produtivo usa outro contrato e secrets permanecem associados na transferência.
- #1721 trata backend/state/exit-code reliability fora do auth do provider.
- #1722 remove o artefato `.tfplan` bruto potencialmente sensível.

gmhelmold:
Cold review identified two pre-existing, non-overlapping follow-ups outside this issue’s provider-auth scope: #1721 (backend/state/exit-code reliability) and #1722 (raw Terraform plan artifact exposure). Both are explicitly tracked and remain blocked on #1720.

gmhelmold:
[...]

PR #2037, opened 2026-09-22
ci(terraform): bootstrap dedicated Cloudflare drift token

gustavomhss:
Refs #1720

## Success criteria
- Owner console installs exactly one dedicated `CF_TERRAFORM_DRIFT_API_TOKEN` with four account reads and two `humangr.com` zone reads.
- One manually confirmed GitHub-hosted audit proves all six GET receipts using that installed token, emits only redacted IDs and scope counts, and dispatches one exact-ref Terraform drift run.

## Invariants
- The workflow cannot create Cloudflare tokens or write GitHub repository secrets.
- No Cloudflare infrastructure mutation and no Terraform apply.
- The bounded audit and Terraform workflows are disabled after the attempt.

## DoD
- Hosted audit succeeds or provides the precise provider/auth denial.
- Terraform drift uses GitHub-hosted runners for this bounded run.


gustavomhss:
Rejected after the final bounded correction. Exact head ec54e7e removes token creation/secret writes, but both disable commands use `|| true`, so an API failure can silently leave workflows enabled. This violates bounded fail-closed disable semantics. No further correction will be attempted in this PR; #1720 remains open.

Issue #1700, opened 2026-09-19
Provision and approve the canonical staging target for load and endurance gates


gmhelmold:
## Classification

- Priority: **P1**
- Type: **operations**
- Delivery impact: **blocks-dependent-work**
- Readiness: **blocked** pending an owner-approved staging boundary and scoped secret provisioning
- Repository: **corelink-server**
- Cosmetic: **no**

## Scope

Create or approve the canonical non-production staging target used by `load-test-nightly.yml` and `endurance-2h-nightly.yml`, and provision only the least-privilege credentials those workflows require. No credential value belongs in this issue, a log, or a workflow body.

## Done when

- [ ] The staging endpoint, ownership, lifecycle, isolation, and cost boundary are documented.
- [ ] Scoped secrets are installed through the approved GitHub/environment path without disclosure.
- [ ] A bounded authenticated readiness probe succeeds from the intended runner fleet.
- [ ] B-029, B-113/endurance and B-113/load-test can execute their own acceptance gates.
- [ ] Rotation and teardown procedures are documented and tested.

gustavomhss:
# #1700 owner checklist after contract preparation

Commit `d4a9e32ef` adds the canonical desired state at `infra/staging/topology.json` and a fail closed verifier. The repository remains intentionally **unprovisioned**; no `terraform apply`, Wrangler deploy, secret write, or live probe was run.

Owner actions required before this issue can close:

- [ ] SRE Lead + Security Lead approve the target, `corelink` runner, 24 hour lease, 2 hour idle teardown, and budgets: USD 25 per load run, USD 50 per endurance run, USD 250 monthly cap.
- [ ] Create exactly the named staging Worker/container, D1, R2, KV, DO, queue/DLQ, service bindings, and route from `infra/staging/topology.json`; capture provider returned IDs and name readback without putting IDs or credentials in git.
- [ ] Render the complete isolated staging Wrangler environment only after readback. Keep `workers_dev = false`, no inherited dev/prod bindings, no cron, and canonical route `https://staging.corelink.humangr.com`.
- [ ] Protect GitHub environment `staging` and bind only these workflow inputs: `K6_TARGET_HOST`, `K6_STAGING_PAT`, `K6_STAGING_STRIPE_WHSEC`, `K6_STAGING_MFA_STUB`, `K6_STAGING_BYOK_CMK_ID`. Values must stay in the approved secret store and out of issue text, logs, plans, and artifacts.
[...]

PR #2040, opened 2026-09-22
ci(staging): add protected three-phase bootstrap lane

gustavomhss:
Refs #1700

## Success criteria

- Render isolated direct Wrangler configurations for exactly the three staging Workers from the frozen topology.
- Bootstrap Workers without public routes, bind only staging inputs, verify secret names, then add only the two canonical routes.
- Run entirely on GitHub-hosted Actions with a protected `staging` environment and no production path.

## Definition of done

- Renderer rejects production targets, secret values, malformed IDs, in-repository output, and routes during quarantine.
- Bootstrap requires exact protected `main`, typed confirmation, distinct control-plane/runtime Cloudflare tokens, every non-derivable staging input, and provider zone ownership before writes.
- Generated internal/admin/erase/PAT material remains worker-local and is never logged or artifacted.
- Readback and ordered manual teardown artifacts contain no secret values.

## Completeness and invariants

- `workers_dev=false`; final routes are restricted to `staging.corelink.humangr.com/*` and the specific pager endpoint, with the generic route published last.
- The root `wrangler.toml` remains free of `[env.staging]`; all rendered configs live under `RUNNER_TEMP`.
- Staging R2 credentials and control-plane credentials are separate from runtime credentials.
- No apply will begin while any external staging input is absent.

## Hosted validation

The PR checks should run the renderer unit suite and static workflow checks. The manual bootstrap remains disabled until merge, all staging inputs, and dual environment approval; then it will be dispatched exactly once on the merged SHA.


gustavomhss:
Rejected after the single bounded correction. Exact head 0cbd8ac correctly adds exact route-set validation, but hosted run 35773434728 fails the renderer against current-main topology because `cloudflare.canonical_origin` is absent. Per the no-loop rule, no second correction will be attempted in this PR; #1700 remains open with environment protection preserved.

PR #2042, opened 2026-09-22
fix(d1): enforce structural B-071 trigger replacement policy

gustavomhss:
Refs #2033 and #1651.

Success criteria:
- 0143 permits exactly the two ordered ADR-0143 DROP plus CREATE trigger pairs.
- Each pair is consumed once; duplicate, unpaired, reordered, renamed, and extra approved-name CREATE statements fail closed.
- Fresh 0142 plus 0143 and historical 0118 plus 0143 replay retain the B-071 accounting, legal-hold, fencing, idempotence, and saturated-byte invariants.

DoD and quality:
- One signed DCO commit based on main d6f63c0.
- Explicit duplicate unpaired approved-name CREATE mutation is covered in both Python B-071 verification and Rust migration property tests.
- Hosted D1 migration validation, B-071 control, rustfmt, and DCO are the only requested checks. No production apply.

Completeness invariant: no generic destructive-DDL or global-find authorization is introduced.


gustavomhss:
Rejected in final replacement review. Exact head 1557d10 passes hosted gates, but both Python and Rust policies remain line-based: `SELECT 1; CREATE TRIGGER <approved-name> ...` on one line bypasses duplicate/unpaired detection, and Rust still globally authorizes the approved name once canonical pairs exist. This violates #2033 statement-level consumed-once invariant. No further correction/replacement loop will be attempted in this campaign pass.

Issue #2044, opened 2026-09-22
[B-097] Bind capacity receipt parser to live Cloudflare schema


gustavomhss:
## Success criteria

The repository-owned parser contract is implemented on `main` by #2070, #2132, and #2170. It accepts only the pinned Wrangler Cloudchamber `GET /accounts/{account}/cloudchamber/me` limits contract, validates finite positive quota values and consistency, and fails closed on malformed/provider-error receipts. Usage and concurrency remain explicitly unavailable; no undocumented field, reservation arithmetic, application count, or unknown response key is treated as a measurement.

This issue is complete only when Cloudflare-authoritative evidence tracked by #2075 establishes the supported account quota, measured-usage, and concurrency signals (exact paths, units, scope, and time semantics), or explicitly states which are unavailable, and any supported signal can be represented by the parser without inference. A successful live receipt requires separate authorization for one bounded protected read.

## Definition of done

- The merged parser and hosted contract checks remain present on canonical `main`.
- #2075 supplies provider-authoritative documentation or support response for each signal, including explicit unavailability where applicable.
- For any supported signal that needs code changes: one signed+DCO PR, exact-head GitHub-hosted focused contract/mutation checks, and one consolidated cold review.
- If a live read is still needed, obtain separate authorization, use one protected GET-only dispatch, retain only the approved redacted receipt, and disable the workflow afterward.
- Root verifies merged-main evidence and closes this issue only when all success criteria are met.

## Completeness criteria

- Quota: `total_vcpu`, `vcpu_per_deployment`, `memory_mib_per_deployment`, and `total_memory_mib` are parsed only from the established `result.limits` contract.
- Usage: explicit unavailable state until a provider-defined path, units, scope, and measurement window are established.
- Concurrency: explicit unavailable state until a provider-defined path and tenant/counting semantics are established.
- Error envelope, missing fields, unknown fields, invalid types, nonfinite/negative/inconsistent values, endpoint drift, and receipt metadata are covered by the fail-closed contract.
- No additional repository-owned parser change is claimed necessary absent new provider evidence.

## Invariants

- No Cloudflare quota or provider mutation.
- No secret, raw response, provider-defined key/path, or unapproved value in logs, artifacts, or Git.
- No extra provider read under this issue; #2075 records that no further dispatch is authorized.
- Preserve the `production-capacity-read` protection and keep the workflow disabled except for a separately authorized bounded read.
[...]

PR #2046, opened 2026-09-22
ci(capacity): capture bounded provider schema topology

gustavomhss:
Refs #2044

## Success criteria
- Adds a protected, main-only, one-GET observation that records only JSON paths/types and the fixed Cloudflare v4 API namespace.
- Retains no raw response, provider values, HTTP-header values, or token.
- Binds the next parser change to an artifact from this exact production observation.

## DoD / completeness
- Signed-off signed commit.
- Hosted focused `issue-1656 capacity` checks must pass.
- After merge, dispatch exactly one protected `run-1656-read-only` observation on `main`; inspect its artifact before any parser binding.

## Invariants / quality
- Fixed `GET /accounts/{account}/containers/me`; no URL/account argument.
- No Cloudflare mutation, deployment, quota change, secret, raw payload, or raw header persisted.
- Bounded topology: 256 nodes, depth 8; arrays are terminal to avoid retaining dynamic identifiers.
- One bounded correction cycle.


gustavomhss:
Final cold verdict: REJECT after the single authorized correction. Hosted focused/DCO/fmt checks are green, but the topology artifact can still retain dynamic identifiers that match the broad lowercase key policy (for example tenant_42 or acct123). That violates the value-free trust boundary. Closing this PR without another review loop; issue #2044 remains open for a different bounded design.

Issue #2049, opened 2026-09-22
[Terraform drift] Bind exact audited SHA through main branch dispatch


gustavomhss:
## Problem

Post-merge #1720 audit run 35774822179 proved all six Cloudflare read receipts, then failed dispatching `terraform-drift.yml` because GitHub Actions rejects raw commit SHA refs (`HTTP 422 No ref found`). Cleanup correctly disabled both workflows.

## Success criteria

- Before dispatch, read authoritative `refs/heads/main` and require its SHA equals the audit run's `GITHUB_SHA`.
- Dispatch `terraform-drift.yml` with `--ref main` only after equality passes.
- Fail closed if main moves, ref read fails, dispatch fails, or disable/readback fails.

## DoD / completeness

- Signed+DCO focused PR with mutation/static cases for moved main, absent ref, raw-SHA regression, dispatch error and cleanup failure.
- No provider apply, token write, or repeated drift dispatch.
- Root performs exactly one post-merge audit dispatch; workflows self-disable.

## Invariants

- The drift run remains bound to the exact audited main SHA.
- No broad ref or PR-head execution.
- One bounded correction cycle.

Refs #1720 and #1722. Evidence: https://github.com/HuGR-dev/corelink-server/actions/runs/35774822179

gustavomhss:
Post-merge evidence from bounded run 35785779614 on `main` at `b6b442b448a3084172d45ab8ad8ac28bdd004669`:

- All six read-only Cloudflare receipts passed.
- The run read authoritative `refs/heads/main`; its SHA matched the audit `GITHUB_SHA`. It issued exactly one Terraform drift dispatch with `--ref main` and `expected_sha`.
- Nested run 35785802004 started on `main` at the same SHA, and its expected-revision guard passed.
- Cleanup readback reports both workflows `disabled_manually`. GitHub returned 403 when asked to disable the audit workflow because it was already inactive; the cleanup step failed closed, leaving the audit run red.
- Terraform init in the nested run stopped before plan because the R2 backend AWS access key ID and secret were unavailable. No apply or additional dispatch occurred.

The exact-SHA dispatch invariant and fail-closed cleanup behavior are evidenced. The separate Terraform backend blocker remains tracked under #1721.

PR #2051, opened 2026-09-22
fix(ci): bind Terraform drift dispatch to audited main SHA

gustavomhss:
Refs #2049

## Success criteria

- Read authoritative `refs/heads/main` through the GitHub API and require its full SHA to equal the audit run's `GITHUB_SHA`.
- Dispatch `terraform-drift.yml` with `--ref main` only after that equality passes.
- Reject absent or moved refs, ref-read failures, dispatch failures, raw-SHA dispatch regressions, and cleanup/readback failures.
- Recheck the expected SHA inside the drift workflow before any Terraform setup or plan.

## Definition of done

- Hosted static and mutation contract covers moved main, absent ref, ref API failure, raw SHA, dispatch error, and cleanup failure.
- Commit is SSH signed and DCO signed.
- No provider apply, token write, repeated dispatch, or local tests/build/lint.

## Completeness and invariants

- The drift run is bound to the exact audited main commit, including the dispatch race window.
- Only `refs/heads/main` is eligible for this audit dispatch.
- The bounded workflow still disables both workflows and verifies `disabled_manually` readback.

## Quality standard

- GitHub-hosted checks are the execution evidence; local work was limited to diff/static inspection.
- One bounded correction cycle is allowed for a concrete hosted failure.


gustavomhss:
Final campaign verdict: REJECT after the implementation's single correction still left the hosted contract red (run 35775947853). The workflow and test disagree on the operand order of the mandatory SHA equality marker. Per the bounded-review policy, this PR is closed without another correction loop. Issue #2049 remains open for a fresh approach.

PR #2059, opened 2026-09-22
feat(audit): exercise nonproduction key custody rotation

### LINKED ISSUE #1794: [B-054] Establish audit signing and link-key custody and rotation evidence
Parent: #1647

## Scope
Establish and exercise write-only custody, rotation, and retention for audit head-signing keys, link keys, and witness receipt keys.

## Acceptance
- Document separate approved custodians and the runtime authorization path for provisioning and rotation.
- Provision a new link key and forward epoch without reusing an id or changing a prior commitment.
- Prove historic keys remain available for the full retention period and that missing or malformed historic material returns `INDETERMINATE`.
- Verify secrets are absent from D1, R2, source, CI output, and evidence artifacts.
- Attach redacted custody and rotation receipts to #1647.

## Invariants
Key bytes never enter durable rows, manifests, logs, or repository files; rotation is forward-only; deleting a historic key cannot yield a clean verification result.

gustavomhss:
Refs #1794

## Success criteria

- The protected, nonproduction GitHub Actions environment proves a forward E1 to E2 key rotation using unique link-key ids and changed commitments.
- The old E1 signing/link material remains available after E2 activation and verifies the historic head and link.
- Missing, malformed, or simulated early-revoked historic material yields fail-closed/INDETERMINATE behavior.
- No secret bytes reach D1, R2, source, CI output, or receipt artifacts.
- A redacted hosted receipt is produced from an approved exact-main run.

## DoD

- Added the gated b054-key-custody-drill workflow, a Rust integration drill over the shipped epoch/keyring/archive primitives, a redaction scanner, and the bounded custody/retention procedure.
- Provisioned four synthetic nonproduction values in the separately protected GitHub environment; names-only readback is recorded in the issue handoff.
- PR CI compiles the secret-gated drill without exposing the values.
- After merge, dispatch once on main, require the independent environment approval, retain the redacted artifact, and close #1794 only after inspecting it.

## Invariants

- Workflow runs only on main; repository permissions are read-only.
- can_admins_bypass=false, self-review is prevented, named reviewers are gmhelmold and gustavomhss, and the allowed branch pattern is exactly main.
- Keys are synthetic and never replace production credentials; no D1/R2/Worker or production mutation occurs.
- E1 is kept while E2 is active. The drill simulates early revocation and malformed/missing history but does not revoke E1 before retention expires.
- Test output is withheld until the redaction scan passes. The artifact contains public commitments and status fields only.

No local tests or builds were run. The environment is intentionally main-only, so the secret-consuming workflow can run after integration; this PR itself cannot access the environment secrets.


gustavomhss:
Final campaign verdict: REJECT. The single bounded correction was consumed and the exact-head hosted rustfmt gate still fails, while DCO, Docs Reality and archive conformance pass. Closing without another correction/review loop. The protected nonproduction custody environment and names-only secrets remain as operational state; issue #1794 stays open for a clean future implementation and hosted rotation receipt.

Issue #1877, opened 2026-09-22
[B-046] Provision and prove an Object-Lock archive adapter


gustavomhss:
### Current disposition

This issue is the external-proof successor for #1646 and #1795. It remains
open and `needs-evidence`.

Canonical `main` at `926600b87e33c1002caeaba994e5361c0dfa500e` already has
the repository-owned seam:

- #1883 merged the provider-neutral `ObjectLockArchiveAdapter` and
  `VerifiedObjectLockArchive` contract. It rejects incomplete negotiation and
  verifies immutable put, retention, legal hold, residency, durable receipt,
  and exact-object delete denial; its R2 adapter fails before an immutable
  write.
- #1895 merged an intentionally inactive AWS S3 Object-Lock template and
  operator gate. It selects neither an AWS target nor a runtime route.
- #1973 makes empty-key and wrong-object denial receipts fail closed.
- #2347 and #2400 completed the distinct AWS adapter recovery: the protected
  verifier binds the same delete-probe principal, bucket, key, version, and
  `s3:DeleteObjectVersion` permission before an `AccessDenied` can count as
  delete-denial evidence.

Those are implementation contracts, not proof of a provider target. PR #2066
was closed after its one allowed correction when the hosted optional-adapter
compile failed with `E0716`. PR #2137 was then closed because caller-supplied
IAM strings were not bound to the credentials that made the delete attempt;
an S3 `AccessDenied` could therefore still have been an IAM denial. Neither
PR, its branch, nor its synthetic receipt is usable provider evidence.

No approved target, OIDC configuration, live receipt, protected audit/digest
record, or independently accepted provider evidence is in the repository. The
current B-046 receipt is `INDETERMINATE`: it failed before bucket creation and
skipped the per-object operation. Existing native R2 Bucket Lock metadata,
lifecycle policies, conditional writes, in-memory fakes, repository tests, and
the inactive AWS template are not Compliance Object Lock or WORM proof.

There is no bounded repository-owned adapter change to make now. The current
hosted contract suite exercises the provider-neutral receipt contract only; it
does not compile or behavior-test a real provider adapter. A future
repository-only adapter proof may be proposed only after a clean baseline
exists and only if GitHub Actions at the exact PR head fully compiles the
provider feature and runs its behavior tests, including failed
negotiation/readback, identity and object/version binding, effective-delete-
authorization binding, and delete success/denial polarity. Such a PR must make
no provider-capability claim and must keep the route unwired until the external
[...]

PR #2066, opened 2026-09-22
feat(audit): add AWS S3 Object Lock archive adapter

gustavomhss:
## Summary

- Adds an opt-in native AWS S3 Object Lock adapter for the provider-neutral audit archive contract.
- Keeps the adapter feature-gated and does not wire an archive production route.
- Narrows the template writer policy to `audit/*`, Compliance retention bounded to the configured term, and legal hold `ON` only; no delete or hold-release permission.
- Records a synthetic-only live probe for AWS account `046797548582` / `us-east-1`: immutable put, retain-until and legal-hold readback, provider-region readback, S3-enforced delete denial, and CloudTrail event references. Bucket/object identifiers and raw logs are excluded.

## Scope and limits

No tenant data was written, no jurisdiction mapping is claimed, and no production route or production approval is included. The locked synthetic object must remain through its retain-until date. The CloudTrail hourly digest covering the recorded probe events was not yet available at capture; digest-chain validation is pending. Therefore this evidence does not close production approval or authorize route wiring.

## Hosted verification

The PR workflow runs the portable Object Lock contract suite, compiles the optional AWS adapter, and validates the inactive Terraform module and IAM invariants. No local tests were run. Merge only after all hosted checks pass.

Refs #1877


gustavomhss:
Final campaign disposition: REJECT after the single bounded correction. The focused checks improved, but hosted Rust compilation still fails with E0716 in read_retention; the unrelated CLI slice gate is also red on this stale base. Closing to prevent another correction/review loop. #1877/#1646 remain open for a clean future implementation.

PR #2079, opened 2026-09-22
feat(capacity): add redacted provider instance receipt

### LINKED ISSUE #2075: [B-097] Obtain documented Cloudflare capacity and concurrency signals
## Parent items

- #2044, which tracks binding the capacity parser to Cloudflare's live schema.
- #1656 / B-097, which remains open until current account quota, all declared reservations, and concurrent tenant occupancy have fresh evidence.

## Evidence from the one authorized read

Protected run [35786516383](https://github.com/HuGR-dev/corelink-server/actions/runs/35786516383) ran on `main@8121050bfd4e65134e136be45b40f8da73788e64` and completed successfully. Its receipt SHA-256 is `8e67d3510e285036ceb249d150048e12ff900728a4c2871f25025afddeb302df`.

The fixed predicates report the API success field as boolean, the errors field as array, and the result field as object. Every allowlisted candidate for account vCPU quota, per-deployment vCPU, memory quota, and usage is missing. The receipt reports 24 unknown object keys, 29 total nodes, and depth 3. The uploaded capture log is empty. Canary scanning found no `tenant_42`, `acct123`, `customer_name`, or token reference. No provider object names or values were retained.

## Missing provider signal

The response does not expose any allowlisted quota or usage fields. Its unknown object members cannot be named from this artifact by design. The public Cloudflare API schema does not document `GET /accounts/{account}/containers/me`, so guessing the 24 suppressed fields would recreate the unsafe discovery path.

Obtain a Cloudflare-authoritative schema/source for a supported read-only signal that establishes:

1. Current account vCPU limit and any per-deployment or memory constraints.
2. Current measured usage, with units, scope, and measurement window.
3. Concurrent active container or tenant occupancy, with units, scope, and measurement window.

If the provider does not expose one of these signals through a supported account endpoint, identify an authoritative documented alternative or explicitly record that the signal is unavailable. No parser may infer it from reservation arithmetic.

## Acceptance

- Link authoritative provider documentation or a provider support response for each available signal and its exact field path/semantics.
- Keep any new observation read-only, fixed-endpoint, secret-free, and limited to compile-time allowlisted paths/types or approved aggregates.
- Bind the parser to the established schema, fail closed on missing or invalid required data, and add contract fixtures before any further read.
- Keep the protected observation workflow disabled unless a separate owner authorization explicitly re-enables one bounded read.

No additional provider dispatch is authorized by this issue.

gustavomhss:
## Summary

- Add a protected, dispatch-only Cloudflare Containers receipt for the five allowlisted production applications.
- Report source timestamp, commit, redacted account scope, documented account ceilings, configured production ceiling, and point-in-time instance state counts with estimated allocation from the declared `basic` shape.
- Retain only aggregate counts; suppress provider stderr and omit raw provider responses, IDs, names, images, locations, versions, and credentials.
- Mark active tenant-region assignments unavailable because instance health and registered tenant state do not establish concurrent workload or tenant-to-instance assignment.
- Extend the hosted fail-closed contract lane with redaction canaries and workflow guard checks.

## Evidence boundary

Cloudflare documents `containers list --json` and `containers instances <APPLICATION_ID> --json`. Published account limits are reported separately from account-specific entitlements, which this read does not verify. Running-instance shape calculations are not provider account usage or billable consumption. Wrangler uses the `containers:write` command permission scope even though this workflow invokes only reads. No provider writes or provider dispatch are part of this PR.

This PR does not close #2075: the provider-authoritative concurrent tenant-region assignment signal and account-specific current quota/usage remain unverified by this supported command route. Do not close the issue until its exact acceptance criteria are met.

## Verification

- `git diff --check` passed.
- Local tests were not run; the existing hosted capacity contract workflow is wired to execute the focused test on this PR.


gustavomhss:
Final campaign disposition: REJECT after the single bounded correction. DCO/docs/fmt and cold semantic review pass, but the hosted credentialless verifier remains red because its command-location contract disagrees with the implementation. Closing to prevent another test-adjustment loop; #2075/#1656 remain open.

PR #2086, opened 2026-09-22
ci(capacity): add bounded Cloudflare instance receipt

gustavomhss:
## Summary

- Add a protected, manual-only Cloudflare Containers receipt that reports published account ceilings separately from the configured five-application ceiling and observed running-instance aggregates.
- Keep account-specific entitlement, account-wide current usage, and tenant concurrency explicitly unavailable.
- Bound each Wrangler command to 90 seconds and 1 MiB streamed stdout; suppress stderr and retain no provider IDs, names, raw responses, or credentials.
- Extend the credentialless hosted capacity contract with redaction, allowlist, ceiling, command, workflow, and oversized-output checks.

## Linked work

- Issues: #2075 and #1656.
- Neither issue is closed by this PR; provider-authoritative tenant concurrency and account-specific quota evidence remain unavailable.

## How tested

- `git diff --check` — passed.
- Local tests and builds were not run as directed.
- Hosted credentialless capacity tests, Docs Reality, DCO, and rustfmt are pending on this PR. The protected provider workflow was not dispatched.

## Risk / rollback

- No deployed service behavior changes. The new provider workflow runs only by manual dispatch from exact `main` after the protected environment and confirmation checks pass. Rollback by reverting this commit removes the receipt workflow, probe, tests, and operator note.

Generated with Codex.


gustavomhss:
Final campaign disposition: REJECT after the single bounded correction. Hosted checks are green, but cold review found the aggregate-only receipt still emits first/last account-ID fragments in account_id_redacted. Closing to prevent a second correction loop. #2075/#1656 remain open; no provider dispatch occurred.

Issue #1653, opened 2026-09-19
[B-083] BYOK: boundary KMS real, fail-closed e ciclo de vida verificável


gmhelmold:
## Success criteria

- The repository-owned B-083 boundary is present on `main`: the shipped image selects `byok-aws-real`; a build with no real provider fails closed; activation constructs the provider and completes `check_access` before its first D1 mutation; and the production revocation detector is constructed with durable collaborators and spawned before listener bind when its explicit operator gate is enabled.
- Closure still requires the owner-authorized, authenticated real-KMS lifecycle receipts tracked in #2165. Repository code or a hosted source-contract run cannot substitute for those receipts.

## Definition of done

- Any repository change is merged to `main`, checked on its exact merged head by the focused GitHub Actions contract, and independently reviewed.
- #2165 records the protected nonproduction runtime, exact image digest and region, least-privilege customer-controlled CMK boundary, and the authenticated lifecycle evidence required there. This issue remains open until that boundary is satisfied.

## Completeness criteria

- The source audit covers provider feature selection, no-provider denial, activation preflight before D1 mutation, Worker-to-container scheduler forwarding, and `run_loop` construction with durable D1 source, fenced status store, and customer notification sink.
- The required lifecycle evidence covers provider access, activation, envelope/CAS wrap and unwrap, revocation and restoration through the production binary, rotation, deletion scheduling, tenant isolation, failure and retry, residency, durable audit delivery, and the measured revocation p99.

## Invariants

- Never close from prose, mocked or local provider behavior, a scheduler flag, source inspection, or unrelated CI. Do not expose or commit credentials, CMK/account/tenant identifiers, ciphertext, plaintext, or provider response bodies.
- The scheduler's explicit enablement remains an owner-controlled deployment gate: enabling it without durable D1, provider credentials, or notification wiring fails before listener bind; leaving it disabled is not evidence that the kill switch operates.

## Quality standards

- Evidence is fresh, reproducible, redacted, and linked to the exact image and runtime boundary. Each completed lifecycle step has a distinct authenticated `audit://` receipt reference and UTC timestamp.
- Use a scoped GitHub-hosted check against the exact head for repository changes, preserve signed+DCO commits, and retain a single independent cold-review verdict for any implementation PR.

### CI Pack

- **Candidate:** External owner-authorized real-KMS lifecycle operation tracked in #2165.
[...]

PR #2107, opened 2026-09-22
fix(byok): require integer probe schema version

gustavomhss:
## Summary

- require the B-083 isolated AWS probe receipt schema version to be the JSON integer `1`
- reject the Python `bool`/`int` type confusion that previously accepted `true`
- add an adversarial mutation covering the forged boolean schema

Refs #1653

## Scope and evidence state

This is a repository-owned parser hardening. The canonical B-083 provider artifact remains a partial isolated provider probe with `closure_eligibility: INELIGIBLE`; it does not claim owner authorization, a protected CoreLink runtime, activation, revocation, rotation, or lifecycle completion.

## Validation

- `git diff --check`
- signed commit with DCO sign-off
- no local tests, builds, or lint run per task constraints
- issue-scoped hosted B-083 contract and adversarial checks requested after opening


gustavomhss:
Superseded by the corrected final-head PR; this PR remains open only at the pre-correction SHA.

PR #2112, opened 2026-09-22
fix(observability): reject malformed B-129 phases

gustavomhss:
## Summary

- Parse each `Server-Timing` item as one complete phase record instead of accepting a valid numeric prefix from `finditer`.
- Reject malformed suffixes and junk before a phase; retain valid trailing parameters such as `desc`.
- Add credentialless rejection cases and a valid-description case to the existing B-129 GitHub PR lane.

## Linked work

- Work item: #1671 / B-129 verifier gap
- Spec ref: #1671 canonical evidence contract
- Closes: none. #1671 remains blocked by #1667 and its production measurement gate.

## Test plan

- [x] `git diff --check` (whitespace/hygiene only; no local tests, build, or lint run)
- [ ] GitHub Actions `issue-1671 B-129 static contract`: `python3 -m unittest -v tests/test_probe_i1671_b129.py`

The probe still requires the five `q*` phases to sum to `wdb`, and `ohop` plus the reported container phases to sum to `origin`. The receipt's total residual remains `(total - wdb - origin) / total`, with median `<10%` as the B-129 measurement threshold. Client `wall_s` is retained as context, not treated as a server phase because it includes network time and this probe reads only a bounded response prefix. This PR hardens the parser feeding those checks; it does not claim all omitted work is attributed or close the production gate.

## Risk

- Customer-visible change: no.
- Schema / API contract change: no.
- Worst case: a malformed phase header is rejected and the diagnostic receipt is withheld, preserving fail-closed evidence.

## Reviewer notes

Please cold-review the two-file parser/test diff. The existing workflow path filter already runs the mutation suite for these files.


gustavomhss:
Rejected after the final corrected-head review: the parser still splits Server-Timing on raw commas and rejects valid quoted descriptions containing commas. The bounded correction is exhausted; issue #1671 remains open.

PR #2120, opened 2026-09-22
fix(gc): require dedicated staging observation workflow

gustavomhss:
## Context

The credentialless B-071 readiness verifier accepted any workflow mentioning `corelink-gc-sweep-production`. On current main that match comes only from the production image build's binary-presence assertion, while the GC dry-run workflow runs the fixture binary. This could hide the missing real staging observation lane.

## What

Require a dedicated manual staging observation workflow and its observation-only/delete-off markers. Add focused regression cases proving an image-build reference is insufficient and the dedicated workflow carries the required mode fences. Run these cases from the existing credentialless readiness workflow.

## How tested

No local tests/build/lint run for this bounded audit. The focused unittest is wired into the existing hosted readiness workflow; its separate step fails on regression. The following readiness audit remains credentialless and continues to report provider readiness blockers.

## Risk and rollback

The change only tightens repository evidence detection. It does not invoke GC or provider APIs. Reverting the commit restores the prior detector.

Refs #1651


gustavomhss:
Closing without merge after the single bounded correction review. Cold review found a remaining false-pass: extra workflow steps can run delete-enabled GC or provider mutations while the required read-only step remains valid. The staging readiness audit is still BLOCKED on unprovisioned provider state, missing readiness receipt/secrets inventory, and missing a dedicated safe observation workflow. Keep #1651 open for runtime evidence.

gustavomhss:
Rejected under the bounded-correction rule: after the sole structural YAML correction, cold review still found an unaudited extra-step path that could invoke deletion-enabled GC/provider mutations. Issue #1651 remains open; no merge or live GC action.

Issue #1721, opened 2026-09-20
[P1] Make regional Terraform drift checks use valid, isolated, locked state


gmhelmold:
# [P1] Make regional Terraform drift checks use valid, isolated, locked state

## Success criteria

The four-region production drift workflow has valid, isolated, locked Terraform state and reports plan outcomes correctly. That repository-owned implementation is already on canonical `main` through PR #1782 (`6cab15f8`). This issue closes only after GitHub-hosted evidence also proves that Cloudflare R2 honors the configured remote-state write/readback, same-key lock contention, normal lock release, and exact temporary-object cleanup against the staging bucket.

## Definition of done

- The merged production implementation remains correct on current `main`.
- A bounded, independently reviewed staging proof workflow is merged by the central owner.
- One exact-main GitHub Actions dispatch using Terraform 1.11.4 and the protected `staging` environment produces a successful redacted receipt for state write/readback, lock contention, release, and cleanup.
- The receipt contains no credentials, Terraform state, or raw plan output.
- The central owner closes this issue after checking the exact run and artifact.

## Completeness criteria

- Production contract: `wnam`, `enam`, `weur`, and `sam` are separate roots with distinct S3 keys; each has an S3 backend and native `use_lockfile`; credentials are runtime-only; missing inputs fail before planning; no generated production stub, `-lock=false`, or apply path exists.
- Plan outcome contract: `0=clean`, `2=drift`, `1/other=error`; `tee` errors remain errors.
- Live staging evidence: run Terraform 1.11.4 against a unique object prefix in `corelink-terraform-staging-state`; read back the written remote state; hold a native lock while a second operation on the same state fails specifically with that lock ID; prove the first operation exits normally and removes the lock; remove and verify absence of only the run’s exact state and lock objects.
- Secret hygiene: the proof uses only the `staging` environment R2 key pair; values never appear in logs, arguments, HCL, or the artifact. Do not upload raw plans, state, or provider logs.
- The previous proof payload in PR #2141 is not on main. Keep #2141 closed; recover its missing capability only in a new current-main change with adversarial checks.

## Invariants

- Never touch production Terraform state or production resources for this proof.
- No Terraform apply/deploy, Cloudflare provider mutation, `-lock=false`, credential promotion to repository scope, or secret-value inspection.
- The proof must fail closed unless dispatched from `main` for the exact expected commit SHA, and must use the protected staging environment.
[...]

PR #2141, opened 2026-09-23
test(terraform): prove R2 native lock contention

gustavomhss:
## Context

Closes the remaining live R2 backend evidence gap in #1721. The existing `staging` GitHub Environment permits only `main`; the first bounded dispatch from this branch was rejected before any steps ran (run 35802413026). This adds the exact-main-capable probe without changing the production drift workflow.

## What

Adds a dispatch-only job to the existing staging provisioning workflow. It uses the `staging` environment's bucket-scoped R2 S3 key pair and a unique run-specific object prefix. Terraform 1.11.4 writes and reads back empty isolated state, then two Terraform plans overlap: the first holds Terraform's native `.tflock` during a slow external data read, while the second must time out on that lock. The first plan must complete and release the lock. The job removes its exact temporary state/lock objects and uploads a secret-free receipt.

## Success criteria and DoD

- Terraform uses the exact pinned 1.11.4 version and actual Cloudflare R2 S3 endpoint.
- State write/readback is verified at a run-unique key in the empty staging-only bucket.
- A second concurrent Terraform operation fails specifically because the first holds Terraform's native lock.
- The lock object disappears after the first plan completes normally; local Terraform state is absent.
- Probe state/lock objects are removed and the uploaded receipt contains only status and identifiers.
- Job contains no apply/deploy command and uses only the `staging` environment credentials.
- Hosted PR checks and DCO pass; after root merge, one exact-main dispatch will provide the provider receipt.

## Invariants and quality

No production state key is used. No secrets are passed as arguments, printed, or uploaded. The only writes are isolated staging state/lock objects under `corelink/issue-1721/<run-id>-<attempt>/`, removed by the job. No duplicate run will be dispatched.

## How tested

No Terraform commands were run locally. Hosted run 35802413026 is infrastructure evidence only: GitHub denied branch deployment because the `staging` Environment policy allows only `main`; both jobs failed before their first step. The workflow was disabled immediately after that one dispatch and remains disabled.

## Risk / rollback

If the probe fails after creating staging state, its EXIT cleanup removes the unique lock and state objects. The entire dispatch-only job can be reverted without affecting the production drift workflow or any regional production state.

Refs #1721.


gustavomhss:
Closed unmerged after cold review found false-success and parallel-job blockers in the live lock proof. The single correction allowance is exhausted. The attempted PR-head dispatch was blocked before job steps by the staging environment's main-only deployment policy, so no live contention proof ran and no success is claimed. Issue #1721 remains open pending a sound, bounded proof.

Issue #2166, opened 2026-09-23
[B-216] Add an authenticated, auditable one-time DSR DLQ redrive authority


gustavomhss:
## Relation

Follow-up to #1678 and rejected PR #2144. #2366 supplies the frozen D1 schema and deployment-secret fences. This issue owns the repository runtime implementation; #1678 remains open for its separate production evidence.

## Success criteria

One operator authenticated by the dedicated `DSR_DLQ_REDRIVE_AUTH_KEY` can submit exactly one eligible terminal DSR DLQ receipt using only its opaque event id. The Worker reconstructs the tenant-bound erasure message from the D1 envelope and emits exactly `_dlq_requeue: 1`.

## Definition of done

- one signed+DCO PR from current `main`, with the frozen `migrations/d1/0145_dsr_dlq_redrive_authority.sql` unchanged;
- exact-head GitHub-hosted B-216 behavioral proof and static mutation proof pass;
- independent cold review approves the exact head;
- root merges serially after revalidating head/base, review, focused proof, and branch policy;
- merge closes this repository implementation issue only after its merged-main state is verified.

## Completeness criteria

- dedicated redrive credential accepts no queue, shared internal, erase, or PagerDuty credential fallback; missing or malformed authority/configuration returns 503 and a wrong bearer returns 401;
- route accepts only `{ "event_id" }` plus bounded operator and approval references; it accepts no caller tenant, legal-hold, salt, payload, provider response, or credential field;
- terminal DLQ ACK requires a captured bounded recovery envelope. Missing, malformed, or unavailable recovery storage retains the queue copy for retry;
- D1 persists only event id, DSR id, tenant id, queued timestamp, legal-hold, requeue count, state, bounded references, and expiry metadata; no raw queue JSON, Clerk identity, salt, request/provider payload, or credential;
- cleanup bounds the envelope to seven days; frozen 0145 has no audit timestamp, so associated audit metadata is removed with its envelope, which is stricter than the 30-day maximum.

## Invariants

- atomic `ready → claimed` claim requires an unexpired envelope and `requeue_count = 0`; a five-minute lease bounds the claim;
- a second atomic `claimed → ambiguous` fence requires both lease and envelope unexpired and occurs before queue submission;
- successful submission transitions `ambiguous → submitted`; any missing envelope, salt failure, fence denial, provider error, or unknown send outcome remains durably non-redriveable and never triggers an automatic second send;
[...]

Issue #1678, opened 2026-09-19
[B-216] DSR erasure dead-letter queue has no consumer and no alerting — exhausted right-to-erasure requests are silently parked forever


gmhelmold:
## Classification

- Priority: **P1**
- Type: **security**
- Delivery impact: **blocks-go-live**
- Readiness: **needs-evidence**
- Repository: **corelink-server**
- Cosmetic: **no**

## Dependency graph

- Blocked by: None.
- Unlocks: None.
- Related, not blocking: None.

**Unblock rule:** Unblocks/finishes on the fresh live or authority evidence named in the acceptance gate; repository prose alone is not proof.

## Scope

B-101 proposal DD-068; this item preserves the distinct finding at `docs/security/2026-06-15-launch-due-diligence-audit.md`, MEDIUM / LOW item 47.

Problem: DSR erasure dead-letter queue has no consumer and no alerting — exhausted right-to-erasure requests are silently parked forever

Evidence: `docs/security/2026-06-15-launch-due-diligence-audit.md#MEDIUM / LOW item 47` — the admitted audit records this exact finding title.

Dependencies: none declared by the B-101 finding; implementation must re-check adjacent controls before closure.

Next action: For DD-068: remediate the finding titled "DSR erasure dead-letter queue has no consumer and no alerting — exhausted right-to-erasure requests are silently parked forever" at docs/security/2026-06-15-launch-due-diligence-audit.md (MEDIUM / LOW item 47); add a focused regression fixture proving the failure mode is closed and fail-closed on missing evidence.

Acceptance/closure evidence: Evidence for DD-068: a code or documented owner decision at the cited source, a regression test that fails before the fix, and a recorded post-fix result linked from this item.

## Done when

- [ ] The current-state evidence satisfies the canonical `B-216` closure contract; any implementation is merged and verified on `main`; the issue links fresh, reproducible evidence; and `BACKLOG.md` is reconciled in the same closing change.
- [ ] Evidence is linked here with secrets and personal data redacted.
- [ ] Any code/config change is merged through its scoped gate and verified on the exact merged tree.
- [ ] No stale branch, PR, temporary worktree, or follow-up debt created by this issue remains unowned.

## Canonical evidence contract

- Source: [`B-216` in `BACKLOG.md`](https://github.com/HuGR-Labs/corelink-server/blob/cca798ff5bc2df660ecf2570ed243eb9775ff3d0/BACKLOG.md)
- Previous backlog state: `parked`
[...]

PR #2174, opened 2026-09-23
feat(dsr): add bounded DLQ redrive authority

gustavomhss:
## Summary

- Add an operator-only DSR DLQ redrive route protected by a dedicated credential with no shared-key fallback.
- Persist a bounded D1 recovery envelope and trigger-backed redacted audit transitions; rebuild the queue message inside the Worker and derive its salt there.
- Add fixed retention cleanup, an operator runbook, behavior-level adversarial tests, and a GitHub-hosted contract lane.

## Linked work

- Work item: #2166
- Related: #1678
- Refs: #2166

## Hosted verification

- Pending GitHub Actions: `pnpm --dir apps/signup-worker exec vitest run tests/dsr_consumer.test.ts tests/dsr_dlq_redrive.test.ts`
- Pending GitHub Actions: `python3 scripts/verify_b215_b223_contracts.py --id B-216 --self-test`
- No local test, build, or lint command was run; this change is explicitly limited to the hosted lane.

## Risk and rollback

- Schema/API contract change: additive D1 migration and new internal route.
- Worst case: an operator redrive remains claimed or ambiguous and requires manual disposition; it cannot automatically send again.
- Rollback: revert this commit. Existing DLQ handling remains intact; post-migration redrive rows are bounded and expire through the scheduled cleanup.

## Review focus

- Verify the D1 trigger makes claim and redacted audit atomic before the queue side effect.
- Verify no queue payload, Clerk ID, salt, provider response, credential, or caller-selected tenant is persisted.
- Verify a duplicate, expired, already-capped, or ambiguous receipt cannot re-enqueue.


gustavomhss:
REJECTED / CLOSED after the single correction budget.

Blocking safety defects remain:
- the claim lease may expire and become ambiguous while submission is still in flight;
- terminal ACK can occur without a retained recovery envelope when recovery storage is unavailable;
- hosted proof does not establish claimed-state behavior or D1 non-persistence;
- the static verifier does not enforce capture-before-ACK ordering;
- the focused redrive lane is red on a stale retry_paging marker.

The only permitted correction was already used on the checkout action pin. This implementation cannot merge. #2166/#1678 remain open for a fresh design that proves atomic claim, recovery-before-ACK, ambiguous-outcome handling and privacy-safe persistence.

PR #2184, opened 2026-09-23
test(reapi): freeze gRPC transport decision

### LINKED ISSUE #2176: [B-113/reapi] Prove Worker → DO → Container gRPC transport transparency
Parent: #2067

## Contract

Before any REAPI service is exposed, prove that `corelink-api.humangr.com` preserves an external HTTP/2 gRPC request through the Worker, Durable Object, and `Container.getTcpPort(50051)` fetcher. The existing proxy only forwards Fetch `Request`/`Response` objects; its comments claim gRPC support but current main has no gRPC service and the container’s Axum dependency enables `http1`, not `http2`.

## Scope

- Define the supported Cloudflare transport behavior from public TLS edge through Worker and DO to the container.
- Enable HTTP/2 at the container only when the selected Axum/Tonic composition has a bounded, tested service shape.
- Preserve `content-type: application/grpc`, binary framing, request metadata including `authorization`, response metadata, status, and `grpc-status` trailers.
- Prove unary and streaming behavior against the deployed path using a non-secret diagnostic service or an equivalent vendor-supported staging proof.

## DoD

- A signed+DCO bounded PR adds a GitHub-hosted protocol/adversarial test that rejects HTTP/1-only composition, dropped authorization metadata, altered gRPC content type, and missing terminal trailers.
- A protected-environment runtime receipt identifies the exact deployed SHA and endpoint, contains no credentials or payloads, and proves the path with a standard gRPC client.
- No Buck2 runtime dispatch, cache write, or endpoint claim is enabled by this issue.

## Invariants

- Do not convert gRPC to REST, HTTP/1, gRPC-Web, a local proxy, or a local cache fallback.
- Do not log, artifact, or persist PAT values.
- Transport uncertainty fails closed: the public ingress remains absent until the exact deployed path is proven.

## Dependency

Blocks all service-mount work in #2067.

gustavomhss:
## Context / why

Closes #2176. The current Worker → Durable Object → `TcpPort.fetch()` bridge is a Fetch composition and has no deployed proof that it preserves HTTP/2 gRPC framing, metadata, trailers, deadlines, or streaming.

Cloudflare's documented socket path is still private beta and requires approved Spectrum/private-beta provisioning. This PR therefore records the bounded fail-closed decision rather than adding a partial ingress.

## What

- Adds a source-grounded transport decision with the provider documentation links, frozen future socket contract, receipt requirements, invariants, and re-open condition.
- Adds a credentialless GitHub-hosted exact-head verifier plus adversarial mutations. It rejects a changed Fetch-bridge claim, removed private-beta condition, missing terminal trailer requirement, credentialed checkout, public gRPC/REAPI route, or premature Worker/DO socket handler.
- Adds no public endpoint, deploy, provider mutation, secret, runtime dispatch, cache write, or REAPI service.

## How tested

- GitHub Actions only: `issue 2176 gRPC transport fail-closed contract` runs the verifier and standard-library adversarial regression at the exact PR head.
- No local tests were run, by contract.

## Risk / rollback

This is a fail-closed source-only decision. Rollback is reverting this one commit; it does not change a deployment or data plane.

## Author checklist

- [x] Commit is SSH-signed and DCO-signed.
- [x] Workflow uses GitHub-hosted execution and a credentialless exact-head checkout.
- [x] No provider setting, secret, runtime dispatch, endpoint, or storage mutation.
- [x] No authentication value, payload, or token-derived material is written to source or artifacts.


gustavomhss:
REJECTED / CLOSED after the single correction budget.

The focused run passed, but final cold review found three unresolved contract blockers:
- no explicit assertion that checked-out HEAD equals the PR head;
- PR-controlled workflow/verifier self-attests instead of using a trusted base-controlled gate;
- the ingress guard rejects only literals and does not reject application/grpc requests before the current Fetch proxy.

The permitted correction was already used on the trailer mutation. This PR cannot prove transport transparency and receives no second correction loop. #2176 remains open.

PR #2190, opened 2026-09-23
feat(dsr): add fail-closed DLQ redrive authority

gustavomhss:
## Summary

Implements #2166 as a bounded, authenticated DSR DLQ redrive authority.

- adds the dedicated operator route, opaque receipt-only request surface, D1 allowlist envelope, redacted transition audit, retention cleanup, and operator runbook;
- makes terminal DLQ ACK fail closed unless a bounded envelope is captured first;
- uses atomic `ready → claimed` and `claimed → ambiguous` pre-send transitions. The latter is a durable send-intent fence, so lease cleanup cannot reopen a receipt while a queue call is in flight;
- reconstructs only the tenant-bound DSR wire payload, derives salt in-worker, and always sets `_dlq_requeue: 1`;
- adds hostile behavior cases and static self-mutations for credentials, caller tenant substitution, duplicate claims, lease/fence denial, durable capture, queue ambiguity, and the persistence allowlist.

## Frozen DoD

One dedicated-authenticated operator action can submit one eligible receipt, once. It must be durable and redacted at `claimed`, pre-send `ambiguous`, and `submitted`; ambiguous outcomes have no automatic retry. The path persists no raw JSON, Clerk identity, salt, secret, provider output, or arbitrary field. #2166 and #1678 remain open.

## Verification

No local tests or checks were run. The PR-specific GitHub Actions lane runs:

```text
pnpm --dir apps/signup-worker exec vitest run tests/dsr_consumer.test.ts tests/dsr_dlq_redrive.test.ts
python3 scripts/verify_b215_b223_contracts.py --id B-216 --self-test
```

Commit `64b5869d0c571bba6dbcce31f9070660988cad87` is signed and includes DCO sign-off. No deployment or merge is requested.


gustavomhss:
Closing unmerged under #2166’s one-correction limit.

The required `redrive-contract` hosted lane remains red at `ff94d95f`: the dedicated-key adversarial mutation replaces the first interface occurrence while the route assertion still sees a later occurrence, so the mutation does not turn the contract red.

The independent cold review also found a safety defect: `prepareDispatch` checks only the claim lease, so a receipt claimed immediately before the seven-day envelope expiry can be sent after its envelope TTL. The fence must also require `expires_at_ms > now` and gain hosted adversarial coverage.

No second correction, deployment, merge, or partial behavior will be attempted. A future clean attempt must address both findings with a new correction budget. #2166 and #1678 remain open.

Issue #2167, opened 2026-09-23
B-071: obtain owner retention decision and provider evidence for bounded GC observation


gustavomhss:
## Classification

- Priority: **P1**
- Type: **operations**
- Delivery impact: **blocks-go-live**
- Readiness: **needs-evidence**
- Repository: **corelink-server**

## Context

Refs #1651 (B-071).

The repository-owned GC observation control is present and fail-closed. Fresh hosted evidence on `main` (`ca41913d2062174071b9e93d049a1938243a2c48`) confirms the legal-hold, tenant/digest/run/region, active-writer, accounting, retry, and idempotent-finalization fences. It is deliberately observation-only: `GC_LIVE_DELETE=true` is rejected and the native observation process has reject-only R2 and audit mutation adapters.

The next boundary requires authority and provider facts that the repository cannot infer:

- staging currently declares `deployment_state: unprovisioned`;
- `evidence/staging/readiness.json` is absent;
- the credentialless verifier cannot prove the required protected staging secret-name set; and
- no owner-approved retention/eviction decision identifies the tenant/region/run scope for the first observation, or separately authorizes a first delete.

## Required decision and evidence

1. Record the approved retention and eviction policy for the selected scope, including treatment of workspace pins, active references, legal/audit holds, grace period, and the rollback/stop condition.
2. Provision and protect an isolated staging target with a D1 binding and a region-matched R2 bucket. Record a redacted provider/readiness receipt at `evidence/staging/readiness.json`; it must include the immutable deployed image digest and the required secret *names* only, never their values.
3. Execute one bounded observation using explicit tenant, region, and run identifiers. The receipt must record candidate count, reclaimable bytes, zero deletes, and the retention/legal-hold/reference/accounting outcomes.
4. Obtain an explicit owner approval before any separately scheduled first live deletion. That later operation must remain bounded to the approved scope and must produce an independently linked receipt.

## Acceptance

- The hosted readiness verifier changes from its expected pre-provisioning BLOCKED state only when the redacted authority and provider evidence exists.
- The observation receipt is reproducible, points to its immutable image digest, and proves zero R2/D1 deletions.
- #1651 links this issue and the receipt. #1651 remains open until its canonical B-071 closure contract is met.
- This issue does not authorize, enable, or perform live deletion.

gustavomhss:
Current-main B-071 / #2167 evidence audit (2026-09-23)

[...]

PR #2196, opened 2026-09-23
fix(gc): require dedicated observation workflow

gustavomhss:
## Summary

- Require the dedicated `.github/workflows/gc-production-observation.yml` path to contain `environment: staging` and invoke the canonical multi-scope collector before the readiness verifier credits an observation workflow.
- Add hosted adversarial contract cases showing that an image-smoke mention of `corelink-gc-sweep-production` fails, a collector command without staging protection fails, and the dedicated staging workflow passes.
- Add that dedicated workflow path to the credentialless readiness workflow trigger. The current credentialless readiness audit remains BLOCKED pending external evidence.

## Linked work

- Work item: B-071 / issue #1651
- Related external dependency: #2167 (not claimed by this PR)
- Closes: none; #1651 remains open for owner/provider evidence.

## Success criteria and DoD

- Only the dedicated observation workflow can satisfy the observation-workflow predicate.
- The hosted adversarial contract passes on this PR's exact head.
- The credentialless readiness audit reports the actual missing staging evidence, without treating an image smoke check as a customer observation.
- No local test, provider call, GC invocation, storage mutation, or delete is performed.
- Diff remains exactly the two files listed below; signed DCO commit is included.

## Invariants

- Observation remains explicitly tenant/region/run scoped through the existing collector.
- The collector continues to force `GC_OBSERVATION_ONLY=true` and `GC_LIVE_DELETE=false`, omit R2 write credentials, and reject nonzero delete count or bytes.
- This verifier change does not authorize or enable live deletion. Zero-delete observation remains the required boundary.

## Scope

- `.github/workflows/issue-1651-gc-live-readiness.yml`
- `scripts/verify_i1651_gc_live_prerequisites.py`

## Test plan

- [ ] Hosted PR run executes `--contract-self-test` on the exact PR head.
- [ ] Hosted credentialless readiness audit reports its expected BLOCKED state until provider readiness exists.
- [x] `git diff --check` is clean.
- Local tests were not run, as requested.

## Risk / rollback

- This only makes the readiness check stricter about what counts as a staging observation. Revert this scoped commit if a correctly protected dedicated workflow cannot satisfy the explicit markers.
- No customer-facing change, schema change, or provider change.


gustavomhss:
Closed unmerged after the terminal cold review found that the matcher can accept required option strings echoed elsewhere in a staging job's run block rather than bound to the collector invocation. The hosted self-test, DCO, and rustfmt passed, but that contract gap remains. Per the bounded review/correction limit, this PR is rejected and will not be merged; #1651 remains open for the owner retention decision and provider evidence in #2167.

Issue #2067, opened 2026-09-22
[B-113/buck2] Add authenticated gRPC REAPI cache ingress


gustavomhss:
## Problem

Issue #2047 cannot reach its intended CoreLink remote cache with the current product surface. Buck2 uses its REAPI client over gRPC. CoreLink currently exposes the relevant cache surface over HTTP/REST and does not mount a gRPC REAPI ingress. The Buck2 starter `.buckconfig` also uses Bazel-style `[remote_cache]` fields and an unrecognized `buck2_re_client.remote_cache_address`; its selected execution platform does not enable remote actions. As a result, workflow run 35778939066 completed the cold build entirely locally (`4 local`, `0 remote`, `remote_cache_hits=0`, `declared_actions=114`) and the warm build reported `0/114`. This is a protocol/configuration gap, not evidence of an unhealthy PAT or a low cache-hit threshold.

Implement the product-side gRPC REAPI cache ingress/bridge required for Buck2 and wire the starter to the real service. The ingress must use CoreLink's existing tenant-scoped storage and auth controls. It must not create a separate cache, silently accept local execution as remote success, or use a runner-local artifact restore as proof of a CoreLink cache hit.

## Scope

- Mount a supported gRPC REAPI endpoint in the deployed CoreLink service topology, with stable `engine_address`, `action_cache_address`, and `cas_address` values for the Buck2 client. The engine address must have explicit semantics; if execution is not supported, fail closed and document the cache-only contract rather than claiming remote execution.
- Implement the Buck2-required REAPI operations for action-cache lookup/update and CAS read/write (including the required ByteStream paths) through the existing tenant-isolated CoreLink CAS/AC storage path. Preserve digest validation, quota, byte accounting, and audit controls.
- Authenticate every gRPC request with the injected CoreLink PAT, enforce tenant and `cache:read` / `cache:write` scopes, and return protocol-correct auth failures. Missing credentials must be rejected before any unauthenticated operation or local-success fallback. Never log or persist token values.
- Configure the starter using Buck2's documented `[buck2_re_client]` endpoint keys and environment-interpolated auth header, plus a checked-in execution platform/`CommandExecutorConfig` that enables the intended remote cache behavior and `remote_enabled` policy. Pin the behavior to the repository's Buck2 release and verify the config is consumed by Buck2.
[...]

PR #2213, opened 2026-09-23
feat: add authenticated REAPI ingress kernel (#2177)

### LINKED ISSUE #2177: [B-113/reapi] Freeze and test the authenticated tenant-scoped ingress kernel
Parent: #2067. Depends on #2176 for public exposure.

## Contract

Implement the small shared kernel used by every future REAPI cache RPC. It authenticates the bearer PAT through `adapter_pat::PatVerifier`, derives the tenant and write capability from the authoritative D1-backed verifier, rejects a non-matching REAPI instance name before storage, and invokes the already-decorated CAS/AC trait objects only.

## Scope

- A typed ingress dependency bundle constructed from the same production-decorated CAS/AC handlers as `routes::build_with_factory_and_byok`; it must not construct a parallel handler set.
- Authentication and error mapping: absent/malformed/invalid credentials → `UNAUTHENTICATED`; verifier backend fault → `UNAVAILABLE`; missing write capability → `PERMISSION_DENIED`.
- Tenant-instance and reserved-namespace rejection before any CAS/AC access.
- One quota/concurrency admission seam whose unavailable outcome fails closed and whose exhausted outcome maps to `RESOURCE_EXHAUSTED`.
- Central SHA-256 digest and REAPI resource-name validation shared by all service modules.

## DoD

- One signed+DCO PR stays within the change budget and contains GitHub-hosted Rust compile, behavior, and adversarial tests for every listed rejection path, including a handler-spy proving denied requests never access storage.
- Cold review checks that the kernel does not retain bearer material, create an in-memory store, or bypass audit, byte accounting, BYOK, tombstone, quota, and tenant prefixes.
- The kernel is not publicly mounted until #2176 and all REAPI service contracts are green.

## Invariants

- Scope and tenant come only from the verified PAT; client metadata cannot broaden either.
- All backend ambiguity fails closed.
- Token strings never reach logs, errors, metrics, artifacts, cache keys, or committed fixtures.

## Dependencies

Blocks the CAS (#2178), ByteStream (#2181), and ActionCache (#2182) service contracts; all remain children of #2067.

gustavomhss:
## Summary

Implements issue #2177 as an authenticated, tenant-scoped REAPI ingress kernel. It verifies bearer PATs through the shared D1-backed `PatVerifier`, binds requests to the verified tenant and write capability, validates SHA-256 and ByteStream blob resource names, and applies one fail-closed quota/concurrency admission seam before returning an admitted context.

The route factory exposes the kernel only when the verifier, quota gate, and D1 tenant-cap resolver are configured. It receives the same decorated CAS and ActionCache trait objects used by the existing router. The kernel remains unmounted pending the #2176 transport proof and later CAS, ByteStream, and ActionCache service contracts. It does not implement those services or their later CAS/ByteStream/ActionCache issues.

## Acceptance and verification

- Focused exact-head workflow checks auth/status ordering, decorated handler reuse, unmounted state, and source mutation cases.
- The hosted Rust gate compiles and runs the kernel behavior tests for rejects, admission ordering, no-storage-on-reject, and resource-name validation.
- No tests, builds, or lints were run locally, per task constraints.
- Commit is SSH-signed and includes the DCO `Signed-off-by` trailer.

Closes #2177 after hosted gates pass; #2067 remains subject to its parent acceptance chain.


gustavomhss:
Closing unmerged per issue owner direction: required exact-head Rust proof is blocked before compilation because the hosted runner has no protoc. See #2177 for the exact setup prerequisite and CI evidence.

PR #2229, opened 2026-09-23
ci(reapi): add trusted gRPC deny gate

### LINKED ISSUE #2227: [B-113/reapi] Bootstrap trusted gRPC transport deny gate
Parent: #2176

## Purpose

Add the protected-base GitHub Actions gate that will evaluate the subsequent #2176 implementation PR without executing PR-controlled verifier code.

## Scope

- `pull_request_target` gate checks out the immutable base verifier and candidate tree separately, with read-only permissions and no persisted credentials.
- Assert the candidate Worker rejects every `application/grpc` media type before any Fetch/DO proxy work; assert a no-store, non-gRPC 503 response and no authorization reflection.
- Assert the source contract documents that a raw socket Worker → DO → Container path, protocol metadata/binary framing/trailers/cancellation proof, and a protected runtime receipt are required before removing the deny.
- Add adversarial fixture mutations.

## Invariants

- Do not execute PR-controlled code in `pull_request_target`.
- No deployment, provider mutation, secret, PAT, local proxy/cache, REST, HTTP/1, or gRPC-Web fallback.
- This gate is an admission control only; it does not claim deployed transport transparency.

## DoD

A signed+DCO gate-only PR with GitHub-hosted evidence and an independent cold review. #2176 stays open.

gustavomhss:
## Context / why

Closes #2227. The prior #2176 contract PR was rejected because its verifier
and workflow ran from PR-controlled source. This change establishes the
protected-base gate before the #2176 implementation is attempted.

## What

- Adds a `pull_request_target` gate with read-only permissions and pinned
  checkout actions.
- Checks out the immutable base verifier and PR candidate separately; it runs
  only the base verifier against candidate source at the exact PR SHA.
- Defines adversarial fixture checks for an early `application/grpc` deny gate,
  no-store response, and required source contract. Candidate code is never
  imported, built, or executed by this workflow.

## How tested

GitHub Actions only: the repository's hosted checks will exercise the
credentialless verifier fixtures. No local test, build, lint, deployment, or
provider action was run.

## Risk / rollback

This is a source-only admission control. Reverting its one commit removes the
future PR gate; it does not create an endpoint, change a Worker, or mutate
provider state.

## Scope

#2176 remains open. A later PR must add the actual early deny implementation
and source contract, then obtain a protected runtime receipt before any public
gRPC transport claim can be made. Refs #2176, #2067.


gustavomhss:
## Terminal verdict: REJECT

The initial cold review identified three blockers. The single permitted
correction removed the path-filter bypass and changed the verifier to compare
the candidate `index_fetch.ts` and deny helper byte-for-byte against a
trusted-base-derived canonical form.

The correction review found a new systemic blocker: the now-unfiltered
`pull_request_target` workflow runs for every PR to `main`, while the trusted
verifier requires the #2176 candidate files that current `main` does not yet
contain. Merging this bootstrap would therefore fail every unrelated PR before
#2176 lands.

The correction budget is exhausted. This PR is closed unmerged. No public
gRPC route, deployment, provider mutation, credential, PAT, cache write, or
runtime claim was made. #2176 and #2227 remain open.

Hosted evidence at `35332b07a1c639cf4d6b2e0832d4f6f7e72c36c1`:

- Passed: DCO, rustfmt, and secrets-drift.
- Python suite failed in the unchanged backlog ledger verifier:
  `docs/campaigns/remediation/backlog-ledger-snapshot-v0003.json: delivered state drifted after last successor`.
- CodeQL was pending when this terminal decision was made.

Issue #2181, opened 2026-09-23
[B-113/reapi] Implement authenticated REAPI ByteStream read and write


gustavomhss:
Parent: #2067. Depends on #2176, #2177, and the frozen #2178 CAS interface.

## Success criteria

- REAPI `ByteStream::Read` and `ByteStream::Write` authenticate through `ReapiIngress`, bind the canonical resource instance to the PAT tenant, and invoke only the decorated tenant-scoped CAS handlers.
- GitHub-hosted exact-PR-head behavior and adversarial evidence proves valid reads and writes, tenant isolation, bounded memory and concurrency, protocol correctness, and fail-closed outcomes.

## Definition of done

- One signed+DCO PR contains the unmounted implementation, focused hosted workflow, behavior/adversarial suite, and source mutation contract.
- The PR has one independent Luna cold review and, if required, one consolidated same-PR correction. Root merges and closes only after the dependency and exact-head gates are satisfied.
- No local CI, self-hosted runner, provider mutation, deployment, public mount, Buck2 configuration, runtime dispatch, or local-cache fallback is added.

## Completeness criteria

- Read parses only canonical tenant-bound blob resources; enforces SHA-256, configured CAS size, offset and limit semantics; performs one decorated read; and emits bounded frames.
- Write parses only canonical tenant-bound upload resources; enforces write scope, ordered offsets, explicit `finish_write`, transport EOF, declared size, SHA-256, configured limit, and exactly one decorated persistence call.
- Invalid, missing, read-only, cross-tenant, malformed, replayed, incomplete, quota-, audit-, and backend-failing requests reach no unauthorized or speculative persistence path.
- `QueryWriteStatus` returns `UNIMPLEMENTED` until durable resume state exists.
- The implementation consumes only frozen #2178/#2177 ingress interfaces. #2176 remains the separate deployment-transport proof and keeps this service unmounted.

## Invariants

- No unbounded aggregation, second store, in-memory upload state, REST conversion, token recording, speculative persistence, or public service mount.
- Authentication, quota, audit, backend, and validation uncertainty fails closed without exposing bearer material or another tenant's object existence.
- The decorated handler remains the sole read/write data plane and a valid write reaches it exactly once.

## Quality standards

- Minimal coherent diff, signed+DCO commits, GitHub-hosted exact-head evidence only, credentialless checkout, one bounded independent cold review, and ledger updates at each material transition.
[...]

PR #2223, opened 2026-09-23
fix(reapi): prove authenticated ingress kernel

gustavomhss:
## Context

Repairs the unmerged #2213 implementation on current `main`. The prior exact-head Rust proof stopped before compiling because the clean hosted runner lacked `protoc`; this PR installs the repository-standard, full-SHA-pinned `arduino/setup-protoc` action before the focused test.

## What

- Adds the authenticated, tenant-scoped REAPI ingress kernel using the shared D1-backed PAT verifier, quota/concurrency admission, tenant-cap resolver, and existing decorated CAS/ActionCache handlers.
- Fails closed for verifier/admission ambiguity, keeps bearer material borrowed only during verification, validates tenant instances, SHA-256 digests, and ByteStream resource names, and leaves the kernel unmounted.
- Adds source mutation checks, an OKF concept with immutable blob anchors, and manifest/index registration for every new container source surface.

## Hosted verification

- Focused workflow: exact PR-head checkout, source mutation tests, and `cargo test --locked -p corelink-server --lib reapi_ingress::tests -- --nocapture` after pinned Protobuf setup.
- No local tests, builds, or lints were run, per the issue constraint.

## Risk and rollback

The kernel is not mounted on any public route. Reverting this PR removes the future-only dependency bundle and its focused workflow.

Closes #2177

PR #2239, opened 2026-09-23
feat(reapi): add authenticated ByteStream contract

gustavomhss:
## Context

Implements #2181 as a stacked child of #2223 at `acbf17bac05bae34eb0d4b809f0b529a39722897`.

## What

- Adds an unmounted authenticated REAPI `ByteStream` service over the #2177 ingress bundle and the existing decorated CAS handlers.
- Validates tenant-bound canonical resources, read ranges, ordered write offsets, finish, declared size, SHA-256, and the 5 MiB CAS ceiling before its single persistence call.
- Streams completed reads in 64 KiB frames and bounds service buffering to three operations within the container read/write budgets.
- Returns `UNIMPLEMENTED` for `QueryWriteStatus`; no upload state or public mount is added.

## Test plan

- GitHub Actions runs the source mutation contract and `cargo test --locked -p corelink-server --lib reapi_bytestream::tests -- --nocapture` on the exact PR head.
- No local build, test, lint, or formatter was run, per #2181.

## Risk and rollback

The service is not routed or instantiated by production composition. Reverting these two commits removes the future-only ByteStream contract and its focused hosted check.

Closes #2181


gustavomhss:
Rejected without merge: the single permitted correction round is exhausted.

Evidence:
- Stacked base: `acbf17bac05bae34eb0d4b809f0b529a39722897` (PR #2223); rejected head: `b4eb860236c0b0eb74ef2e6d847f948731b209d9`.
- Independent cold re-review verdict: **FIX-FIRST**. The remaining P1 is `cargo fmt --all --check` (module order and test wrapping). The remaining P2 is that `now_or_never()` can miss a replay arriving immediately after `finish_write`; the service needs a bounded terminal drain or an EOF-required policy plus a delayed-replay regression test.
- Hosted `cargo fmt` run [35820474915](https://github.com/HuGR-dev/corelink-server/actions/runs/35820474915/job/107051030280) failed with those formatting findings. DCO passed.
- The focused ingress run [35820474795](https://github.com/HuGR-dev/corelink-server/actions/runs/35820474795/job/107051029839) failed before compilation because `arduino/setup-protoc` encountered a GitHub API rate limit. The focused ByteStream run [35820474825](https://github.com/HuGR-dev/corelink-server/actions/runs/35820474825/job/107051030094) was still pending at rejection.

No local build, test, lint, or formatter was run. The branch and worktree will be removed.

Issue #2182, opened 2026-09-23
[B-113/reapi] Implement REAPI ActionCache and cache-only capabilities


gustavomhss:
Parent: #2067. Depends on #2176 and #2177. Sibling of CAS and ByteStream service contracts.

## Success criteria

- Implement only REAPI `ActionCache.GetActionResult`, `ActionCache.UpdateActionResult`, and authenticated `Capabilities.GetCapabilities` using frozen #2178 ingress interface and already-decorated tenant-scoped Action Cache handlers.
- Preserve complete vendored REAPI `ActionResult` wire shape required by pinned Buck2 client; do not truncate unknown output, metadata, or symlink fields.
- Validate action digest, nested output digests, inline output bytes, and maximum serialized result size before update.
- Delegate lookup/update to decorated AC handlers, retaining immutable/deduplicated updates, audit, BYOK, quota, byte accounting, and tenant namespace enforcement.
- Return `NOT_FOUND` for miss, `ALREADY_EXISTS` for divergent immutable overwrite, and fail closed on storage/audit ambiguity.
- Advertise SHA-256 and ActionCache update support only; do not advertise or emulate remote execution.

## Definition of done

- One signed+DCO bounded PR with GitHub-hosted Rust compile plus behavior and adversarial tests for authorization, tenant mismatch, action-result round trip, malformed nested data, quota/audit fault, miss disclosure, immutable conflict, and no-execution capability.
- Tests prove every successful lookup/update reaches decorated AC path and every rejected request reaches neither handler.
- Exact-head hosted proof is green and one independent cold review returns `APPROVE` or resolved `FIX-FIRST`.

## Completeness criteria

- No public mount, Buck2 configuration, runtime dispatch, or local-cache fallback is added.
- No alternate AC store, partial protobuf persistence, REST adapter, execution service, or token disclosure is introduced.
- #2178-owned interface files are not edited; missing seam is reported to central with exact need.

## Invariants

- Cache remains tenant-scoped and cache-only.
- Rejected requests reach neither decorated handler.
- Successful operations retain decorated-handler audit, BYOK, quota, byte-accounting, immutability, and namespace controls.
- Capabilities never imply remote execution support.

## Quality standards

- Work in isolated worktree from current `origin/main` or explicitly frozen dependency head.
- Run tests, builds, lints, mutations, and verification only in GitHub Actions on GitHub-hosted runners.
- Preserve frozen #2178 interface; no broad refactor or unrelated baseline repair.
[...]

Issue #2183, opened 2026-09-23
[B-113/reapi] Mount the complete cache-only ingress and wire Buck2 acceptance


gustavomhss:
Parent: #2067. Depends on #2176, #2177, #2178, #2181, and #2182.

## Contract

Mount the complete authenticated REAPI cache ingress only after its transport, common kernel, CAS, ByteStream, and ActionCache contracts are individually green. Then wire the pinned Buck2 starter to the supported public endpoint and retain the exact-head runtime acceptance route for #2047.

## Scope

- Compose the complete tonic service set with the existing Axum application on the one container listener, preserving `/_health`, all HTTP routes, graceful shutdown, HTTP/2 behavior, and a fail-closed unmounted state when ingress dependencies are absent.
- Set all Buck2 endpoint keys to the verified TLS gRPC endpoint, use environment-only bearer metadata, exact repository/SHA namespace provenance, and cache-only execution policy.
- Add one focused GitHub-hosted workflow that compiles and runs behavior/adversarial tests on the exact PR head; do not modify unrelated workflows or spend a correction cycle moving CI triggers.
- Keep the one protected post-merge #2047 dispatch as the sole runtime warm-cache admission; it may run only after this issue is merged and root authorizes it.

## DoD

- One signed+DCO bounded integration PR receives a cold review and exact-head GitHub-hosted compile, behavior, and adversarial evidence.
- Hosted tests prove Axum routes remain live, all REAPI services are reachable over HTTP/2, cache-only capabilities reject execution, missing/invalid PATs fail before storage, and no local cache fallback or endpoint substitution exists.
- The post-merge #2047 receipt proves real authenticated CAS/AC writes in cold build and reads in a clean warm build; it records non-secret operation counts and exact repository/SHA provenance, meets the >=80% threshold, and only then admits the benchmark JSON.

## Invariants

- No public mount before every dependency has green hosted evidence.
- No execution capability, REST/gRPC-Web conversion, second store, runner-local artifact restore, threshold weakening, or credential disclosure.
- No merge without a separate cold review; at most one correction for a concrete review or CI defect.

## Dependencies

Unblocks runtime acceptance in #2047 only after the full #2067 chain is merged.

gustavomhss:
## Metadata normalization ledger (2026-09-24)

- Normalized the acceptance headings; integration scope, dependency list, cache-only boundary, ownership, labels, and #2047 runtime gate are unchanged.
[...]

PR #2244, opened 2026-09-23
feat(reapi): add authenticated ActionCache contract

gustavomhss:
## Context

Implements the bounded #2182 child contract stacked on #2223 at `acbf17bac05bae34eb0d4b809f0b529a39722897`. The service remains unmounted; public transport composition stays blocked on #2176 and #2183.

## What

- Restores the REAPI ActionCache RPC and complete pinned `ActionResult` output, metadata, and symlink wire fields.
- Adds authenticated tenant-scoped lookup/update over the exact production-decorated ActionCache handlers, with digest, inline-byte, nested-digest, and serialized-size validation.
- Returns `NOT_FOUND` for misses, `ALREADY_EXISTS` for divergent immutable writes, and fail-closed errors for admission, audit, and backend ambiguity.
- Adds authenticated SHA-256 ActionCache capabilities with no execution capability.

## Hosted verification

- Exact-head workflow installs pinned Protobuf tooling, runs mutation checks, then runs `cargo test --locked -p corelink-server --lib reapi_action_cache::tests -- --nocapture`.
- No local tests, builds, or linters were run, per the issue constraint.

## Risk and rollback

No gRPC route, runner configuration, execution service, or storage implementation is added. Revert this single commit to remove the future-only service contract.

Closes #2182
Refs #2177


gustavomhss:
Superseded by #2268. The fork PR ref remained pinned to its earlier head after updates, so the final signed+DCO correction was moved to a canonical repository branch and hosted verification is running on #2268.

